PR #7, opened 2016-02-23
Fixing retrieval of original mbuf in ToDPDKDevice.

### LINKED ISSUE #6: [PATCH] Fixing retrieval of original mbuf in ToDPDKDevice
Hi again,
This time (I believe) I managed to fix a new issue, a possible patch follows.

The patch should be `git am`-able on top of your master. Also, both [this patch](https://github.com/davidek/fastclick/commit/539c14c2a62678df3494a816bba167785a845b44) and  [another verbose version with prints around](https://github.com/davidek/fastclick/commit/2b904a303adb9270803e30b10312ec652ae1025c) are available at my fork (at the `fix-todpdk-get_mbuf` and `fix-todpdk-get_mbuf-DEBUG` branches respectively).

The weired thing is that, while explicitly trying to access the data in the wrong packet from click causes a segfault, it appears that when DPDK tried to forward it, it would pick an actual `mbuf` containing an old packet (causing the old packet to be sent again).

As usual, I am compiling against DPDK `2.2.0` (perhaps they changed some internals of the `struct mbuf`?), compiling with `./configure --enable-multithread --disable-linuxmodule --enable-intel-cpu --enable-user-multithread --verbose CFLAGS="-g -O0" CXXFLAGS="-g -std=gnu++11 -O0" --disable-dynamic-linking --enable-poll --enable-bound-port-transfer --with-netmap=no --enable-zerocopy --disable-dpdk-pools --enable-dpdk --disable-batch`. I was running with `-c 0x1 -n 1` and sending packets one at a time (funny thing, even the first packet incurred in the issue and it would replicate a packet from the previous session. This drove me mad). The conf file I was using:
```
FromDPDKDevice(0, MINQUEUES 2, MAXTHREADS 64, THREADOFFSET 0, NUMA no, RSS_AGGREGATE yes)
  -> CheckIPHeader(OFFSET 14, VERBOSE yes) // needed for some prints in the *-DEBUG branch
  -> ToDPDKDevice(0);
```

I realize using `destructor_argument` is not the cleanest thing in the world, but it seems reasonable to me that if a Packet was allocated from an mbuf, then `_destructor_argument` must be set, and I believe nothing in the path has a good reason to override it (while at times `_destructor` may get set to the no-op).

For the record, here is the output generated with the prints in the `-DEBUG` branch:
```
-- 24.0.32.69 > 1.2.3.4
dump mbuf at 0x0x7f68b55c5000, phys=0, buf_len=0
  pkt_len=0, ol_flags=0, nb_segs=0, in_port=0
dump mbuf at 0x0x7f68b55c4f80, phys=7f55c5000, buf_len=2176
  pkt_len=73, ol_flags=2, nb_segs=1, in_port=0
  segment at 0x0x7f68b55c4f80, data=0x0x7f68b55c5080, data_len=73
  Dump data at [0x7f68b55c5080], len=20
00000000: 54 75 D0 C4 4F FF EC F4 BB E6 1E 14 08 00 45 00 | Tu..O.........E.
00000010: 00 3B 00 01 |  |  |  |  |  |  |  |  |  |  |  |  | .;..
-- 24.0.32.69 > 1.2.3.4
``` 

```diff
From 539c14c2a62678df3494a816bba167785a845b44 Mon Sep 17 00:00:00 2001
From: Davide Kirchner <davide.kirchner@yahoo.it>
Date: Tue, 23 Feb 2016 10:15:15 +0100
Subject: [PATCH] Fixing retrieval of original mbuf in ToDPDKDevice

---
 elements/userlevel/todpdkdevice.cc | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/elements/userlevel/todpdkdevice.cc b/elements/userlevel/todpdkdevice.cc
index c9ce65e..9597398 100644
--- a/elements/userlevel/todpdkdevice.cc
+++ b/elements/userlevel/todpdkdevice.cc
@@ -128,7 +128,7 @@ inline struct rte_mbuf* get_mbuf(Packet* p, bool create=true) {
     if (likely(DPDKDevice::is_dpdk_packet(p) && !p->shared())) {
         /* If the packet is an unshared DPDK packet, we can send
          *  the mbuf as it to DPDK*/
-        mbuf = (struct rte_mbuf *) (p->buffer() - sizeof(struct rte_mbuf));
+        mbuf = (struct rte_mbuf *) p->destructor_argument();
         rte_pktmbuf_pkt_len(mbuf) = p->length();
         rte_pktmbuf_data_len(mbuf) = p->length();
         static_cast<WritablePacket*>(p)->set_buffer(0);
@@ -180,7 +180,7 @@ void ToDPDKDevice::flush_internal_queue(InternalQueue &iqueue) {
             // The sub_burst wraps around the ring
             sub_burst = _iqueue_size - iqueue.index;
         r = rte_eth_tx_burst(_port_id, queue_for_thread_begin(), &iqueue.pkts[iqueue.index],
-                             iqueue.nr_pending);
+                             sub_burst);
 
         iqueue.nr_pending -= r;
         iqueue.index += r;
-- 
1.9.1


```
EDIT: minor format change -- I've discovered how to colorize the diff

davidek:
Closes #6

### DIFF
--- a/elements/userlevel/todpdkdevice.cc
+++ b/elements/userlevel/todpdkdevice.cc
@@ -128,7 +128,8 @@ inline struct rte_mbuf* get_mbuf(Packet* p, bool create=true) {
     if (likely(DPDKDevice::is_dpdk_packet(p) && !p->shared())) {
         /* If the packet is an unshared DPDK packet, we can send
          *  the mbuf as it to DPDK*/
-        mbuf = (struct rte_mbuf *) (p->buffer() - sizeof(struct rte_mbuf));
+        mbuf = (struct rte_mbuf *) p->destructor_argument();
+        assert(mbuf);  // all DPDK packets should have destructor_argument set
         rte_pktmbuf_pkt_len(mbuf) = p->length();
         rte_pktmbuf_data_len(mbuf) = p->length();
         static_cast<WritablePacket*>(p)->set_buffer(0);
@@ -180,7 +181,7 @@ void ToDPDKDevice::flush_internal_queue(InternalQueue &iqueue) {
             // The sub_burst wraps around the ring
             sub_burst = _iqueue_size - iqueue.index;
         r = rte_eth_tx_burst(_port_id, queue_for_thread_begin(), &iqueue.pkts[iqueue.index],
-                             iqueue.nr_pending);
+                             sub_burst);
 
         iqueue.nr_pending -= r;
         iqueue.index += r;


davidek:
With respect to the original patch, I added a defensive assertion so that if (it is compiled with assertions enabled and) the assumption turns out to be wrong it doesnt segfault.

tbarbette:
Thought of a problem just in time... If the packet is shared and we have the clone, the destructor argument is not set, only the destructor argument of the p->data_packet()...

Is using mbuf = (struct rte_mbuf *) (p->buffer() - sizeof(struct rte_mbuf)); really a problem?

It should be constant...

davidek:
I also thought of this possibility, but it appears to me the current implementation of `clone()` copies the `_destructor_argument` in the `memcpy` operation and does not overwrite it (why should it?).
Nevertheless, it might be worth to check `if (!mbuf && p->data_packet()) {...}`. Do you want me to ad it to the PR?

Yes, `(p->buffer() - sizeof(struct rte_mbuf))` caused bad things to happen on my system (my best bet would be that the difference is in the dpdk version? I'm using `2.2.0`): if you look at the output of the debug prints I copied over in #6 you'll see that the addresses are indeed different on my instance (which in turn caused wrong packets to be sent around by dpdk...)

tbarbette:
Mhh yes, only the _destructor is resetted to 0 in clone() just after the memcpy, I thought it was both...

That's at least the less unclean for now.

Thanks !

tbarbette:
No the assert is enough, if not too much already... This is in a very fast path, called for every packet. At 40G every instruction counts... We should just think it through, rely on tests, and have the fewer runtime test as possible....

davidek:
Indeed, I figured copiling for performance would require `-DNDEBUG`? (Perhaps a line in the README mentioning this would be worth?)

tbarbette:
Never used it. Does it change a lot? Disabling all assert is not a good idea either. Some are needed to prevent mis-configuration from the user. In our present case, I meant it is possible to completely think of all cases, go through all possible changes of in _destructor_argument, and be sure that the code is correct and we can rely on destructor_argument() without asserting after.

But there is some assert in some place which check for example that the ip header is set, and that only relies on the user having placed a MarkIPHeader or CheckIPHeader before...

Issue #3, opened 2016-02-19
ToDPDKDevice: Avoiding extra copy when sending shared() Packets


davidek:
While inspecting the stack traces for #2, it appears that the issue arises when `ToDPDKDevice` tries to _copy_ a packet's data, after the packet has been `clone()`d (i.e., when `shared()` is true).

I was wondering whether that copy is needed: given that (a) the packet cache is thread-private and (b) after the clone, the packet content is completely read-only, would it be possible to avoid this extra copy? In the standalone version we are using `rte_pktmbuf_refcnt_update` to prevent the mbuf from being released after sending, and I'd like to try and implement the same strategy.

While thinking about the possibilities we have to replicate that behaviour, I realized of that `--(dis|en)able-dpdk-pools` compilation flag: reading around the code it seems that enabling it causes a `Packet` to be just a casted `mbuf` , but I didn't understand what effects this has overall, nor why in the README you suggest to keep it disabled for best performance (nor where annotations are stored): would you mind elaborating a little?
I'm particularly interested in understanding whether it would be best to implement this dont-copy-even-if-cloned strategy to work with or without the flag enabled.

Thanks a lot,
Davide

tbarbette:
Hi,

For the dpdk-pools performances : 
The dpdk-pools allows to completly get rid of the Packet object from Click. In DPDK the descriptor data (length, L3 layer offset, user-metada, ...) is at the beginning of the buffer, so there is only one "memory object" per packet. So we replaced every "packet descriptor accessor" of the Click Packet object with a wrapper around the DPDK equivalent function, and it works (while maybe a little untested for some functions). We thought that removing an unnecessary structure, and  an unnecessary copy of packet length, metadata, ... would enhance the speed, but it didn't.

The problem is that to keep all Click functionalities inside the DPDK packet metadata (including annotations), we need two cacheline at the beginning of the mbuf, so we had two cache miss for every packet. But when we use Click Packets linked to the mbuf data itself, but re-using always the same (burst-size) Packet objects, we have only one cache miss, as packets are recycled in Click inside the ToDPDKDevice, and reused at next run in the FromDPDKDevice, therefore there is absolutely no cache miss with the Packet objects.

This results would probably go with low-size cache processors, but ours have 20MB of cache, so the Packet objects stay there.

For the clone, you touch a big problem we have with staying "click fully-compatible" or going for optimized speed. 

[...]

PR #8, opened 2016-02-23
Added missing timer-thread mapping in ToDPDKDevice

davidek:
As discussed in #3


davidek:
You meant like this? In my case it indeed works as expected now (well, in my case it even worked by plugging in the thread number directly...).
Before the fix, I experienced lost packets because `push_packet` would enqueue the incoming packet in the local iqueue, but then the timer would always be triggered in thead zero and always polled the other queue.

davidek:
(you were faster than me :D )

tbarbette:
Your first version was good.  iqueues are per-thread, not per-queue, so "i" had to be directly used... I commited a patch to revert that in master.

PR #12, opened 2016-05-24
Batching support for some elements

gkatsikas:
Please review my modifications and feel free to accept or reject them :)

### DIFF
--- a/elements/ethernet/arpresponder.cc
+++ b/elements/ethernet/arpresponder.cc
@@ -3,7 +3,11 @@
  * arpresponder.{cc,hh} -- element that responds to ARP queries
  * Robert Morris
  *
+ * Computational batching support
+ * by Georgios Katsikas
+ *
  * Copyright (c) 1999-2000 Massachusetts Institute of Technology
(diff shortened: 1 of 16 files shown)


tbarbette:
Thanks !

I should implement a kind of SimpleBatchElement which wraps simple_action like you do by default. I cannot do it by default on all batchelements because some of them are not "functional", the may push packets (not batch) without returning a packet and break things internally.

In the meantime, we'll continue like that... Thanks for the effort !

If you change elements based on push() and not simple_action(), note that I recently had to change push to push_packet in all BatchElement, to differentiate an old push() from a vanilla Element to a push of a BatchElement in non-batch mode. So BatchElement should implement push_packet(int port, Packet* p) and push_batch(int port, PacketBatch *) while vanilla element must implement push(int port, Packet * p). This is not very documented so...

gkatsikas:
I noticed that change with push_batch.
I am currently working on the IPSec library, adding batching support to some of the encrypt/decrypt elements.
I already use your push_batch function there, in a different way than simple_action of course ;)

I will be back soon, with more updates.

Regards,
Georgios

PR #13, opened 2016-06-07
Batching support for IPSec library, some analysis elements, and some …

gkatsikas:
…basic IP elements. Also, a new element is implemented for timestamping TCP/UDP/ICMP packets.

Please review and feel free to accept or reject my PR.


tbarbette:
Thanks! But it seems some checks fail. It seems that you override push in directiplookup, but as discussed earlier, you should name that function push_packet in a BatchElement.

tbarbette:
Regarding StoreTimeSeqRecord, it's not very the "Click way" to implement inside the element the parsing of headers. It should be more like any IP element which would complaint the network or transport header not being set but do nothing about it. It should be the user's problem to correctly set it.
After the transport, the problem is that packets do not have "payload" headers. So that part have to be handled internally depending on the type of packets.

This is interesting, my colleague @cffs implemented kind of the same element and should be reverting it soon. I imagine you also have the other end to compute the difference...

gkatsikas:
Oh sorry for this, there is something I have changed in batchelement.hh in order for these elements to work.
In line 66 and 198, I omitted the "final" keyword of the push methods because they cause these compilation errors.
Can you quickly comment these 2 keywords and see if the tests pass?

gkatsikas:
Yes, StoreTimeSeqRecord is an ugly way of timestamping packets from different protocols but it substantially simplifies the application.
Nowadays, most of the traffic we test is IP-based using either TCP or UDP, hence an application can use this element as follows:

FromDevice($iface)
                -> StoreTimeSeqRecord(OFFSET 14, DELTA false)
                -> ... Processing ...
                -> StoreTimeSeqRecord(OFFSET 14, DELTA true)
                -> ToDump($pcapFile);

In that sense you don't need a classifier if you know that IP traffic is handled by your application.

tbarbette:
I could comment the final keyword but I won't^^
This is to avoid a nasty error when mixing batch compatible and incompatible elements. If an old element just before your element calls push, it will call your version of push, but the one I want to be called which is BatchElement::push which will handle correctly batch-reconstruction if it needs to, and call push_packet if it doesn't ;) Doing as you say will cause problem with --enable-bound-port-transfer as your push function will think it calls push for the next element, but in practice it will be push_batch and the next element will follow unsetted "next" annotation...

tbarbette:
Yes I understand your point of view on IP classification. And yes it leads to easier configuration. But if we follow that way we'll end up with a Router element which does all what Click allows to do in multiple elements...

gkatsikas:
Ok, I got it.
Then let me work around these problems and come back to you with some fixes.

tbarbette:
Well that's a hell of a job... Thanks ! Is it reasonably tested? I mean it works?
At least it creates tests problems : https://travis-ci.org/tbarbette/fastclick/jobs/138353305 do you know about that? It fails with and without batching.

Could you just make a new PR with all DPDK-related stuff? 
There is more discussion about that. I guess you could make the rings work finally? What was the problem? Just updated DPDK?
No way to have only options on To/From DPDKDevice instead of a new element? I imagine you would if you could...
Also I have a piece of shitty code that you copy/pasted, I'll update it in ToDPDKDevice in a few hours, that's the part about filling the internal_queue in batching : instead of wrapping around I add the packets before the oldest if we are at the end of the buffer. That was just to try the impact and forgot to revert. No matter the performance gain, this introduce inacceptable reordering.
So please remove *dpdk* from this PR, and create a new one for that. I also just re-organized this week (From|To)(DPDK|Netmap)Device elements, It will help with conflicts...

gkatsikas:
Hi,

For some reason I mistakenly pushed to master. I didn't mean to do that since I know it is hard to review all these changes at once.
The ring-based communication between multiple DPDK processes works just fine.
One reason I created different elements is to keep the code of the original To/FromDPDKDevice elements short.
Also my new classes do not inherit QueueDevice because the task/thread manager of QueueDevice causes some issues when I try to link many processes.

I have a proposal to keep the repository in tact. Can you simply remove my PR and then I will come back to you with smaller PR requests. Is it ok?
I can isolate the DPDK-based stuff as you requested.

tbarbette:
I'll remove this PR then. I'm delighted to review the sub-PR ;)

PR #14, opened 2016-06-17
More batching in IP library

gkatsikas:


tbarbette:
Is there any reason why in iproutetable you do not use the macro?

Also all inheriting elements of iproutetable will make belive they support batching, but it will in fact call their ancestor's push_batch function... I guess putting in_batch_mode to BATCH_MODE_NO in the constructor of all those inheriting elements should solve the problem until someone actually implements the inheriting push_batch function. Could you do that? Let's put a "//TODO : Remove this when push_batch() will actually be implemented."

Thanks !

tbarbette:
I think your change in README was already done in you other commit, and that prevent merging... Maybe you shoud rebase on master?

PR #15, opened 2016-06-21
A patch with a new pair of DPDK-based I/O elements that read/write from/to shared ring buffers.

gkatsikas:


tbarbette:
Okay, that's great ! Thanks !

It would be even better to write a testie test. But it would need some changes on the way DPDK arguments are passed. Your elements introduce a way to actually try some DPDK stuff on automatic tests without having a real device (such as on travis). I think I have an old branch with some code. 

With OVS, you have to pass dpdk arguments with --dpdk [dpdk stuffs], then they do some tricks to contain EAL parser in those bounds. This would allow Click compiled in dpdk mode to stay compatible with tests. I'll come back with that.

gkatsikas:
Right, feel free to modify these elements, the code can be certainly improved.
I didn't have time to do a testie that's why I quickly uploaded two working Click configurations, one primary and one secondary.
One thing to notice is that you should run these processes in different cores (of the same NUMA zone) if you want the get the best performance. If you attempt to run them on the same core, the PMDs interfere and you will very likely get a race condition.

PR #20, opened 2016-12-17
Fix QUEUE/N_QUEUE bug

lauuuuuura:
The code is not working properly if QUEUE != 0.


tbarbette:
Yep. Thanks !

The Queue/N_queue feature was made not a long time ago to allow manual pinning of queues and cores to input devices, but is not much tested. One goal of FastClick was to abstract queue and cores, sharing all available cores among devices respecting numa nodes, not needing to use StaticThreadSched element or Queue/Nqueues parameters but it was requested for specific cases.

PR #22, opened 2017-01-07
Fix in dpdk.mk. FastClick compilation is now successful when using t…

gkatsikas:
…he DPDK Mellanox MLX4 and MLX5 drivers


tbarbette:
No authorship problem on reverting this into mainline Click I guess?

gkatsikas:
Sure, no problem, I just pushed from my forked repo :)

PR #23, opened 2017-02-02
RatedSource supports batching

gkatsikas:


tbarbette:
Thanks for the PR ! Please separate commits for style and real changes next time (I've done it in local).

However in run_task, why are you calling _tb.set_full() instead of _tb.refill() like in the per-packet case?

gkatsikas:
Sorry for mixing style and functional changes, I will keep your advice in mind next time.
Good point, during the debugging process I was experimenting with the token bucket functions and I found that both functions do the work for me.
I will modify this line though to have 1-1 mapping between the batching and non-batching cases.

tbarbette:
I merged a splitted version of your commit in space/changes (kept you as author for the change part and mentionned you in the style part). I also had to check that the burst size of the batch would not extend beyond _limit, which was causing some error in testies. Thanks !

PR #25, opened 2017-02-19
Stlye changes in EtherEncap

gkatsikas:


tbarbette:
Thanks ! However Eddie is rewriting tabs to spaces on some files and it seems from now on that the convention is "4 spaces" instead of tabs. I updated your commit accordingly to use spaces.

gkatsikas:
Oh, good to know this, thanks!

PR #26, opened 2017-02-20
Batching support for FastUDPFlows

gkatsikas:
If you approve this, I can also push a similar patch for TCP and UDPv6

### DIFF
--- a/elements/tcpudp/fastudpflows.cc
+++ b/elements/tcpudp/fastudpflows.cc
@@ -2,7 +2,11 @@
  * fastudpflows.{cc,hh} -- fast udp flow source, a benchmark tool
  * Benjie Chen
  *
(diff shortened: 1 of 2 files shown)


tbarbette:
Thanks ! I used that one but using Replay* elements so it is only in a preload phase, not needing especially performances. Generally, if you use this as a pktgen, FastUDPFlows->MultiReplayUnqueue will yield much more throughput.

gkatsikas:
Yes exactly, I followed your example configuration in conf/fastclick ;)
I just wanted to get rid of the annoying message that FastUDPFlows will un-batch and re-batch packets, so I decided to make it compatible with FastClick

tbarbette:
Ok nice ! Good to know that people read stuffs :p

PR #35, opened 2017-04-26
Thread-safe IPRewriters. Most of them support batching (except for ICMP)

gkatsikas:


gkatsikas:
I don't understand why the tests fail, could you please shed some light?
I don't see any problems in my tests.

gkatsikas:
I see consistency problems in the IPRewriter tests. Could this happen because the tests were written with a single core context in mind?

elements/ip/iprewriterbase.cc
```diff
@@ -68,17 +71,41 @@ IPMapper::rewrite_flowid(IPRewriterInput *, const IPFlowID &, IPFlowID &,
...
+    //click_chatter("[IPRewriterBase]: Allocated %d heaps", _mem_units_no);
 }
 
 IPRewriterBase::~IPRewriterBase()
 {
-    if (_heap)
-	_heap->unuse();
+    if (_heap) {
+        for (unsigned i=0; i<_mem_units_no; i++) {
+            //_heap[i]->unuse();
```
tbarbette: Normal?

tbarbette:
I'm looking a little at this. Just as a general thought, the per_thread class is made for the per_thread duplication and protects against false sharing of cache lines.

elements/ip/iprewriterbase.cc
gkatsikas: If you uncomment this line, the router does not stop properly, it crashes.

gkatsikas:
I suggest to move this patch into a separate branch for now. I can work on it after June 19.
I do not see a better way to allow IPRewriter to run in a multi-core context, without completely changing the logic of the element. Right now, this family of elements is built with a particular structure in mind.
Feel free to propose alternatives and improvements.

tbarbette:
I'm okay with the way you do it, just proposed to use the per_thread class which automatize some of the things you do to duplicate variables per_thread.
I catched one bug, the timer is initialized on only one core, so the per-thread structure will only be read for the core where the timer runs. Also there using a per-thread<Timer> should solve that (and looping for initialization among all cores). But I'm not sure this is the (only?) bug causing the error in the tests, as the Click tests are single threaded it should not create any problem without going MT for real. I partially corrected that, I'll make a branch tomorrow as you suggested.

tbarbette:
The branch is there : https://github.com/tbarbette/fastclick/tree/gkatsikas-thread_safe_iprewr 
It is now fixeds, _timeouts pointers problems. However I'd still like to rework some things and check for performance impact over the single-core version.

gkatsikas:
There is a slight performance degradation compared to the single core version, I tested the code a year ago. However, this element alone can go at 40 Gbps without decreasing the throughput ;)

tbarbette:
This has been merged with a mention to your work, and most of your commits for legacy.
For relevance most of the problems were output().push() left in some path, breaking batching.

PR #40, opened 2017-07-03
SetipDscp

gkatsikas:


tbarbette:
simple_action_batch should not be implemented if you have both pull_batch and push_batch (which is the best case as simple_action(_batch) add a virtual call).

ALso the pull_batch seems wrong, it pulls a batch and return it untouched.

gkatsikas:
I will remove the simple_action_batch, but why pull_batch is wrong? The batch pointer is filled by the pull() for every packet. This is how other elements implement the pull_batch too.

tbarbette:
It creates a batch by pulling but do nothing with it. smaction(p); should be called for every packet pulled.

gkatsikas:
But smaction is called by pull(), which is called by pull_batch()

tbarbette:
Oh my bad. Okay then.

gkatsikas:
So after my last commit, it is ready to merge :)
Thanks!

tbarbette:
I'm waiting for travis. Yep in my head it was input.pull and not call to the local pull

gkatsikas:
I see that the tests are complaining about elements/tcpudp/fastudpsrcip6.cc:18, which has nothing to do with the modified element.
Why is this the case?

tbarbette:
Master is broken, I'm fixing that

PR #44, opened 2017-07-07
Update kerneltun.cc

IurmanJ:
KernelTun get_spawning_threads overridden


tbarbette:
Thanks for the PR !

However you need to have all files included in a single PR or it will break testing. That also allows better review. You can however make a PR with multiple commits at once. Just push all the commits in one branch on your fork and then create a pull request.

PR #48, opened 2017-09-07
NIC vendor and driver information in DPDKDevice

gkatsikas:


gkatsikas:
One problem about this patch is that the FromDPDKDevice handler that advertises the vendor ID returns a meaningless uint16_t as a string. It would be nice to map this integer to the vendor's name. I haven't found how to do so in DPDK (check here http://dpdk.org/doc/api-1.6/rte__pci__dev__ids_8h.html).

gkatsikas:
DPDK provides no way to retrieve NIC vendor information so I found this way as a quick solution.
If one wants to do it better, he should link FastClick with libpci.

tbarbette:
Seems the current travis fail is my fault.. I'll fix that and try again.

gkatsikas:
I rebased this commit according to the current master and there is no conflict.
If it passes the tests then maybe you can merge right away :)

tbarbette:
It seems the break of ./test/userlevel/dpdk.testie is legit. Can you try on your side or do you want me to check that?

gkatsikas:
I added some more rigorous checks for acquiring vendor information. If this does not solve the issue, feel free to further investigate the issue

gkatsikas:
DPDK v2.2 might have different fields in struct rte_eth_dev_info causing another crash.
I will look into it, hopefully this will be the ultimate fix.

gkatsikas:
DPDK v2.2.0 works ok for me.
I emulated the testie by running:

sudo bin/click --dpdk --no-huge -m 128MB -c 0x3 -n 1 --vdev=eth_ring0 -v -- testie_code.conf

I got the intended behavior:
Initializing DPDK
Starting pktgen
   1 | 41
   1 | 41

tbarbette:
Thanks !

PR #49, opened 2017-09-21
EtherRewrite can pull and push in batch mode

gkatsikas:


elements/ethernet/etherrewrite.cc
```diff
@@ -71,12 +63,39 @@ Packet *
...
-	return smaction(p);
+       return smaction(p);
     else
-	return 0;
+       return 0;
+}
+
+#if HAVE_BATCH
+PacketBatch *
+EtherRewrite::simple_action_batch(PacketBatch * batch)
```
tbarbette: Either simple_action(_batch) or the combination of pull(_batch), push_(batch) should be implemented but not the 3. Having pull+push is slightly faster than simple_action, so I would suggest removing simple action.

elements/ethernet/etherrewrite.cc
```diff
@@ -71,12 +63,39 @@ Packet *
...
+{
+    EXECUTE_FOR_EACH_PACKET(smaction, batch);
+    return batch;
 }
 
 void
+EtherRewrite::push_batch(int p, PacketBatch *batch)
+{
+    FOR_EACH_PACKET(batch, p) {
+        smaction(p);
```
tbarbette: smaction may return another packet, so the batch would be broken.
You must use EXECUTE_FOR_EACH_PACKET(fnt,batch) that will repair the batch if the packet changes

elements/ethernet/etherrewrite.cc
```diff
@@ -71,12 +63,39 @@ Packet *
...
+    }
+    output(0).push_batch(batch);
+}
+
+PacketBatch *
+EtherRewrite::pull_batch(int p, unsigned max)
+{
+    PacketBatch *batch = input_pull_batch(0, max);
+    FOR_EACH_PACKET(batch, p)
+        smaction(p);
```
tbarbette: Same.

elements/ethernet/etherrewrite.hh
```diff
@@ -58,15 +57,15 @@ class EtherRewrite : public BatchElement { public:
     void add_handlers() CLICK_COLD;
 
     inline Packet *smaction(Packet *);
-
     Packet *pull(int);
+    void push(int, Packet *);
 
 #if HAVE_BATCH
     PacketBatch* simple_action_batch(PacketBatch *);
```
tbarbette: Delete

tbarbette:
Merged manually

PR #59, opened 2018-03-12
New socket element for monitoring purposes

gkatsikas:
This element acts as a UDP or TCP client that reports
a list of handlers' values and names to a remote UDP or TCP
server. Currently, this element is useful for element handlers
that contain arithmetic values (e.g., latency, throughput, etc).
The structure of each message is <timestamp, handler value, handler name>.


tbarbette:
I would have put the handler name first, but OK ^^

gkatsikas:
This is not very hard to change. I will consider updating this element in the near future :)

PR #62, opened 2018-03-14
Prevent potential access to unallocated memory in queue dev

gkatsikas:
This might occur when another element has been configured prior
to a queue device and attempts to call the thread_for_queue
function

### DIFF
--- a/elements/analysis/timestampdiff.cc
+++ b/elements/analysis/timestampdiff.cc
@@ -55,15 +55,20 @@ int TimestampDiff::configure(Vector<String> &conf, ErrorHandler *errh)
             .complete() < 0)
         return -1;
 
-    if (get_passing_threads().weight() > 1 && !_limit) {
-        return errh->error("TimestampDiff is only thread safe if N is set");
-    }
-
     if ((_rt = static_cast<RecordTimestamp*>(e->cast("RecordTimestamp"))) == 0)
         return errh->error("RECORDER must be a valid RecordTimestamp element");
 
     if (_limit) {
         _delays.resize(_limit, 0);
+    }
+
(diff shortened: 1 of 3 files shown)


tbarbette:
I would actually prefer this to be an assert() and set the dependency right, eg set the initialization of the caller after FromDPDKDevice. Here we would miss important bug. Also this function is called a lot in the fast path. It would be  better to do the check in get_spawning_threads. Other uses of _thread... are sure to be after initialization. What Element was the caller in your case?

gkatsikas:
In my test, the problem is likely caused by TimestampDiff during configure().
Specifically, in this function TimestampDiff calls get_passing_threads() which invokes thread_for_queue() in queuedevice before initialize_tasks() is called (to allocate the _queue_to_thread memory).

If I just replace my commit with an assert, is it ok for now?

gkatsikas:
This patch fixes the issue between FromDPDKDevice and TimestampDiff.
If other elements try to access thread information before being initialized (e.g., at the configuration phase), then the assertion will be fired.

tbarbette:
Still it will add multiple check for each packet in the fast path. I would prefer the check being done outsite of thread_for_queue, in get_spawning_threads for example. Thread_for_queue is sure to be inialized after all elements initialization, so no need to check that in the fast path.

gkatsikas:
Any additional insights on this?

tbarbette:
Thanks. And you solve the bug then...

gkatsikas:
Can you also merge all the latest changes with Metron?

tbarbette:
Done

gkatsikas:
Super!

PR #64, opened 2018-03-15
RecordTimestamp takes numbered packets from input element

gkatsikas:
This allows to use the desired element if more than one
are present in a Click configuration


tbarbette:
I've got many configs in many place that depends on this. Could you make it non mandatory, and if not given use the singleton?

gkatsikas:
Yes, I was also worried about this fact.. I am going to update asap

PR #70, opened 2018-03-28
DPDK Batch support for UDPIP6Encap

kthkaya:
DPDK Batch support for UDPIP6Encap


tbarbette:
The commit name is wrong as this has nothing to do with DPDK ;) But thanks, I'll merge as it !

PR #82, opened 2018-05-03
Add FORCE_SHRINK parameter to the EnsureHeadroom element

piotrjurkiewicz:
`EnsureHeadroom(FORCE_SHRINK 1)` can be put after each element generating packets on slow path (`ICMPError`, ARP speakers) in order to explicitly remove headroom from a packet. That way, warnings about data move can be avoided in `ToNetmapDevice` for these packets. So, when such a warning occurs, it will be certain that it comes from some fast path packet.

### DIFF
--- a/elements/standard/ensureheadroom.cc
+++ b/elements/standard/ensureheadroom.cc
@@ -22,7 +22,7 @@
 CLICK_DECLS
 
 EnsureHeadroom::EnsureHeadroom()
-    :  _headroom(Packet::default_headroom)
+    :  _headroom(Packet::default_headroom), _force(false)
 {
 
 }
@@ -32,6 +32,7 @@ EnsureHeadroom::configure(Vector<String> &conf, ErrorHandler *errh)
 {
     if (Args(conf, this, errh)
 	.read_p("HEADROOM", _headroom)
+	.read_p("FORCE_SHRINK", _force)
 	.complete() < 0)
 	return -1;
     return 0;
@@ -40,7 +41,7 @@ EnsureHeadroom::configure(Vector<String> &conf, ErrorHandler *errh)
 inline Packet* EnsureHeadroom::smaction(Packet* p) {
 	int length = p->length();
 	Packet* q;
-	if (p->headroom() < _headroom) {
+	if (p->headroom() < _headroom || _force) {
 		q = p->shift_data(_headroom - p->headroom());
 	}	else
 		q = p;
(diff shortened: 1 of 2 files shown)


tbarbette:
Shouldn't it be FORCE alone? It may force "expand" and not only "shrink", right?

For the problem at hand. It doubt it will actually change something in term of performances, you'll replace a memcpy by a memmove. May seem better in term of cache but actually it will lead to less netmap buffer being used (as it uses FIFO), so the destination may already be in cache and therefore memcpy be faster. Did you test it?
I don't remember the message but if it's for each packet, of course it should be silenced after some times.

The real problem is that Netmap does not support headroom, as you point out. Actually, I did modify Netmap some time ago to change the descriptor "ptr" field by an union, with an u16 being the headroom for both TX and RX, and also an u32 to get the RSS hash in RX. I had implemented support for ixgbe, but i40e should be very easy too. The address of the buffer is recomputed every time anyway with netmap, so this did not impact performance. I can look for that version if you want. However the authors will never merge that patch I think, because it collides with the pointer when using VALE. And adding a new field to the descriptor will hit the ring size badly... Though I think it is a very interesting feature... If you encap or decap packets, it makes Netmap unusable...

piotrjurkiewicz:
> Shouldn't it be FORCE alone? It may force "expand" and not only "shrink", right?

It doesn't make sense to force expand, since it happens always (when packet headroom is shorter then specified of course).

This checks whether packet headroom should be expanded (packet headroom is shorter than requested). If it is true, argument passed to `p->shift_data` is positive.

    if (p->headroom() < _headroom) {
 		q = p->shift_data(_headroom - p->headroom());

Hence, adding `|| _force` covers the opposite case, when packet headroom is longer than requested. In such a case, argument passed to `p->shift_data` is negative, so packet headroom is shrinked.

So. to sum up, force may only force shrink, so I called it FORCE_SHRINK.

> For the problem at hand. It doubt it will actually change something in term of performances, you'll replace a memcpy by a memmove.

I am aware of that. This parameter is to be used solely for diagnostic purposes, to silence data shift warnings for packets traversing them.

Warnings in `ToNetmapDevice` do not specify where is the packet originating from. For example in my configuration I have three packet sources:

1. `FromNetmapDevice` fast path -- warnings about data shifting for these packets would indicate some bug within FastClick/netmap
2. `KernelTap` for interfacing with routing daemon -- by default creates packets with headroom, but can be set to 0 with `HEADROOM` parameter
3. `ICMPError`, ARP generators, etc. -- these elements always create packets with excessive headroom (it is hardcoded)

When I was getting warnings, I didn't know which packets cause them. After explicitly putting `EnsureHeadroom(FORCE_SHRINK 1)` after elements in group 3 and still getting the messages I discovered that I forgot to set `HEADROOM 0` in KernelTap (I wasn't even aware that `KernelTap` creates packet with headroom).

tbarbette:
Ok I only overlooked the dhrink issue. Makes sense. Thanks.

PR #89, opened 2018-07-16
Preliminary support for DPDK 18.08

gkatsikas:
This patch is based on DPDK 18.08-rc1

### DIFF
--- a/lib/dpdkdevice.cc
+++ b/lib/dpdkdevice.cc
@@ -335,7 +335,9 @@ int DPDKDevice::initialize_device(ErrorHandler *errh)
     tx_conf.tx_thresh.pthresh = TX_PTHRESH;
     tx_conf.tx_thresh.hthresh = TX_HTHRESH;
     tx_conf.tx_thresh.wthresh = TX_WTHRESH;
+#if RTE_VERSION <= RTE_VERSION_NUM(18,05,0,0)
     tx_conf.txq_flags |= ETH_TXQ_FLAGS_NOMULTSEGS | ETH_TXQ_FLAGS_NOOFFLOADS;
+#endif
 
     int numa_node = DPDKDevice::get_port_numa_node(port_id);
     for (unsigned i = 0; i < info.rx_queues.size(); ++i) {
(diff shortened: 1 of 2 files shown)


gkatsikas:
Do not merge yet. Let this pull request be alive through the whole 18.08 life-cycle

gkatsikas:
While compilation works, I see a run-time issue with 18.08-rc2 and Intel 82599 NICs.
When I start a test app, I get:

Initializing DPDK
Ethdev port_id=0 invalid rss_hf: 0x3afbc, valid value: 0x38d34
master.click:117: While initializing ‘io0/td :: ToDPDKDevice’:
  Cannot initialize DPDK port 0 with 8 RX and 8 TX queues
  Error 1 : Operation not permitted
Router could not be initialized!

Mellanox MLX5 works fine :)

tbarbette:
I would bet the old igb_uio from DPDK 18.05 was loaded and this would be your problem. Monday thoughts.

gkatsikas:
Could be the case, I will check again..

gkatsikas:
18.08-rc3 does not introduce additional dependencies and does not require any fixes compared to 18.0-rc2. However, I still get the same error when using Intel NICs. I am pretty sure I load the new (correct driver. There is something else wrong here.. Mellanox still works fine!

tbarbette:
As a reminder, use this as default : 
"preferred Rx/Tx parameters.

Rather than applications providing explicit Rx and Tx parameters such as queue and burst sizes, they can request that the EAL instead uses preferred values provided by the PMD, falling back to defaults within the EAL if the PMD does not provide any. The provision of such tuned values now includes the ixgbe PMD."

tbarbette:
I pushed to master directly without really wanting it. I fixed it, see the commit for more details. The problem was the default RSS parameters that are now verified for unsupported mode.

gkatsikas:
No problem at all. I will do my tests tomorrow, including Metron. Hope it works just fine

gkatsikas:
FastClick master with DPDK 18.08 works for me!
It is time to merge the changes with metron as 18.08 brings several enhancements on the Flow API.

PR #96, opened 2018-08-15
Parameter for DPDK pause frame mode

tbarbette:
And also a few type correction to avoid warnings.

### DIFF
--- a/elements/userlevel/fromdpdkdevice.cc
+++ b/elements/userlevel/fromdpdkdevice.cc
@@ -55,7 +55,6 @@ int FromDPDKDevice::configure(Vector<String> &conf, ErrorHandler *errh)
     bool has_mac = false;
     bool has_mtu = false;
     FlowControlMode fc_mode(FC_UNSET);
-    String mode = "";
 
     if (parse(Args(conf, this, errh)
         .read_mp("PORT", dev))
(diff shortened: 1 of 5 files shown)


elements/userlevel/fromdpdkdevice.cc
```diff
@@ -54,13 +54,16 @@ int FromDPDKDevice::configure(Vector<String> &conf, ErrorHandler *errh)
     uint16_t mtu = 0;
     bool has_mac = false;
     bool has_mtu = false;
+    FlowControlMode fc_mode(FC_UNSET);
+    String mode = "";
```
gkatsikas: This variable seems to be unused

lib/dpdkdevice.cc
```diff
@@ -416,6 +416,26 @@ int DPDKDevice::initialize_device(ErrorHandler *errh)
         }
     }
 
+    if (info.init_fc_mode != FC_UNSET) {
+        struct rte_eth_fc_conf conf;
+        ret = rte_eth_dev_flow_ctrl_get(port_id, &conf);
+        if (ret != 0)
+            return errh->error("Could not get flow control status !");
+        switch (info.init_fc_mode) {
```
gkatsikas: [Detail] You could allocate a static Map with string as key and the enum as a value.
this way you can simply get from that map

PR #97, opened 2018-08-16
TimestampDiff

tbarbette:
Allows to give an index from which the average/mean/max/... should be computed. This index is a packet number, maybe read using "index". By default, index is 0, the old behaviour.

This ultimately allows to display the change of latency over time.

### DIFF
--- a/test/analysis/latency-begin.testie
+++ b/test/analysis/latency-begin.testie
@@ -1,0 +1,48 @@
+%info
+RecordTimestamp ecosystem
+
+Test the RecordTimestamp ecosystem (TimestampDiff, NumberPacket, ...)
+
+%script
+click -j 1 CONFIG
+
+%file IN1
+!data ts_sec len payload
+1 60 A
+2 60 B
+3 60 C
+4 60 D
+5 60 E
+10 60 F
+15 60 1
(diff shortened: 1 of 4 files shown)


test/analysis/latency-begin.testie
```diff
@@ -0,0 +1,48 @@
...
+-> diff :: TimestampDiff(RECORDER record)
+-> Counter(COUNT_CALL 5 double.run)
+-> Counter(COUNT_CALL 10 finished.run)
+-> Discard;
+
+double :: Script(TYPE PASSIVE, write du.delay 5)
+finished :: Script(TYPE PASSIVE, read diff.average, read diff.average 0, read diff.average 5, read counter.count, stop)
+
+%expect stderr
+diff.average:
```
gkatsikas: You could update the testie to cover other important handlers such as median, min, and man

tbarbette: In this test there is some randomness so we could not test the value. latency-01 and -02  could be modified to do so.

tbarbette: I would like to introduce predictable delay but that would require another element to do so.

PR #98, opened 2018-08-17
click: Add --simtick options

tbarbette:
Allows to change the amount of tick to warp in simtime mode

This allows to force us scale to enable some testing


tbarbette:
This actually works, but I've got other patches incoming to fix travis false positives.

PR #99, opened 2018-08-17
Timewarp test: relax constraint and more robust

tbarbette:
The timewrap test fails from time to time on travis, which is very slow.
The time to launch the process makes the delay longer than expected, so
let's relax the constraints.

In addition, substract from the real time the time it takes to launch an
empty click. This allows to be less error-proned regarding the time
a process takes to execute, eg on travis.


tbarbette:
Also maybe @bcronje  may review this?

PR #105, opened 2018-08-27
Batching helpers

tbarbette:
3 helper element to easily build batch-compatible without downgrading performances. The explanation is actually already in the wiki : https://github.com/tbarbette/fastclick/wiki/Batching

And then PaintRR which uses one of them.

### DIFF
--- a/elements/standard/paintrr.cc
+++ b/elements/standard/paintrr.cc
@@ -1,9 +1,8 @@
 /*
- * paintswitch.{cc,hh} -- element routes packets to one output of several
- * Douglas S. J. De Couto.  Based on Switch element by Eddie Kohler
+ * paintrr.{cc,hh} -- element routes packets to one output of several
+ * Tom Barbette.  Based on PaintSwitch element
  *
- * Copyright (c) 2002 MIT
- * Copyright (c) 2008 Meraki, Inc.
+ * Copyright (c) 2018 KTH Institute of Technology
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
@@ -38,7 +37,7 @@ PaintRR::configure(Vector<String> &conf, ErrorHandler *errh)
     return 0;
 }
 
-int
+inline int
 PaintRR::classify(Packet *p)
 {
     return static_cast<int>(p->anno_u8(_anno)) % noutputs();
--- a/elements/standard/paintrr.cc
+++ b/elements/standard/paintrr.cc
@@ -1,0 +1,49 @@
+/*
+ * paintswitch.{cc,hh} -- element routes packets to one output of several
+ * Douglas S. J. De Couto.  Based on Switch element by Eddie Kohler
+ *
+ * Copyright (c) 2002 MIT
+ * Copyright (c) 2008 Meraki, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, subject to the conditions
+ * listed in the Click LICENSE file. These conditions include: you must
+ * preserve this copyright notice, and you cannot mention the copyright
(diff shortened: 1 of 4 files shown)


include/click/batchelement.hh
```diff
@@ -131,5 +131,127 @@ class BatchElement : public Element { public:
...
+
+#if HAVE_BATCH
+    void push_batch(int port, PacketBatch* head) override final {
+        head = T::simple_action_batch(head);
+        if (head)
+            output_push_batch(port,head);
+    }
+
+    PacketBatch* pull_batch(int port, unsigned max) override final {
+        PacketBatch* head = input_pull_batch(port,max);
```
gkatsikas: assert max, unless input_pull_batch() does it

include/click/batchelement.hh
```diff
@@ -131,5 +131,127 @@ class BatchElement : public Element { public:
...
+        if (head) {
```
gkatsikas: Be consistent in parentheses for single statement conditions. Either you always use the parentheses or you never do it.

include/click/batchelement.hh
```diff
@@ -131,5 +131,127 @@ class BatchElement : public Element { public:
...
+
+    private:
+        SimpleBatchElement(){};
+        friend T;
+
+};
+
+
+/**
+ * Batch helper element
```
gkatsikas: you need a full stop at the end of this sentence.

include/click/batchelement.hh
```diff
@@ -131,5 +131,127 @@ class BatchElement : public Element { public:
...
+ * extend SimpleElement<T> where T is the element itself.
+ * It also avoids the virtual call, so actually a vanilla element using
+ * this version even without batching will run faster.
+ *
+ * The inherited element cannot be extended further because of CRTP !
+ */
+template <typename T>
+class SimpleElement : public SimpleBatchElement<T> { public:
+
+    inline PacketBatch* _sm_action_batch(int port, PacketBatch* batch) {
```
gkatsikas: Also be consistent with brackets. Should you always open a bracket at the same line with the function name or always put it in the next line?

include/click/batchelement.hh
```diff
@@ -131,5 +131,127 @@ class BatchElement : public Element { public:
...
+};
+
+
+/**
+ * Batch helper element.
+ *
+ * Allows the user to simply implement simple_action and simple_action_batch
+ *  but relying on CRTP instead of virtual, removing one virtual call.
+ *
+ * The inherited element cannot be extended further because of CRTP !
```
gkatsikas: What countermeasures do we have for this restriction?
If we convert all Click elements to this template, how can we leverage inheritance?
Maybe using template parameters, but this makes the code base more complicated.

elements/standard/paintrr.cc
```diff
@@ -0,0 +1,49 @@
+/*
+ * paintswitch.{cc,hh} -- element routes packets to one output of several
+ * Douglas S. J. De Couto.  Based on Switch element by Eddie Kohler
+ *
+ * Copyright (c) 2002 MIT
+ * Copyright (c) 2008 Meraki, Inc.
```
gkatsikas: Change the authorship if this is your example.

include/click/batchelement.hh
tbarbette: I generally do on the same line.

include/click/batchelement.hh
tbarbette: This is a fast easy way to make an element compatible. If one wants inheritance he can copy paste the code in the CRTP. Or extend the base template to pass the inherited and the final inheritance. But in any case if someone wants to do something complex he'll implement push* and pull* himself...

include/click/batchelement.hh
tbarbette: I prefer not to add an assert here. Each element should not ask for 0 packet, that's a coding error. We can't add run-time checks for every of thoses.

PR #106, opened 2018-08-28
xstats bug fixed

tbarbette:


gkatsikas:
This commit is related to FastClick, thus it should also be applied to master

tbarbette:
The bug was introduce by a metron feature I think.

gkatsikas:
See my pull request #108

PR #108, opened 2018-08-28
Fix return value of xstats handler

gkatsikas:
Also, some style fixes to comply with space-based
tabs of 4 characters


tbarbette:
The function finishes by a return 0 so the value was actually okay. but I'll take the style fixes  :p

gkatsikas:
I missed that, it was hidden under the #endif :p

gkatsikas:
I amended the commit message and fixed the duplicate return, you can merge directly

PR #109, opened 2018-08-29
ReplayUnqueue : Allow to replay according to ts

tbarbette:
If the TIMING parameter is > 0, ReplayUnqueue will replay
according to the packet timestamp. TIMING is an integer and if
bigger than 1 will define the acceleration speed. A value of 2
will replay packets twice faster, ...


tbarbette:
It is not ready actually...

PR #113, opened 2018-09-21
If using DPDK, allocate memory from DPDK mem

tbarbette:
A little memory improvement. Ensure head of packet is on a boundary, and if using DPDK use the dpdk system for packet allocation.

Allows to simply use huge pages and cache alignment (mem is allocated on
boundary, and headroom is changed to be a multiple of cacheline size)


bcronje:
Also happy with the increase in default_headroom

PR #114, opened 2018-09-21
EtherEncap batching helper

gkatsikas:
Batching is done using the helpers and not the old-style way


gkatsikas:
EnsureEther does exactly what I wanted, so I will push a new PR which introduces batching support for that element

PR #115, opened 2018-09-21
Improvment to the set of replay elements

tbarbette:


elements/analysis/replay.cc
```diff
@@ -139,6 +150,9 @@ Replay::configure(Vector<String> &conf, ErrorHandler *errh)
     .complete() < 0)
         return -1;
 
+    if (_timing > 0)
```
gkatsikas: You don't you put this parameter directly in ReplayUnqueue?

elements/analysis/timestampdiff.cc
```diff
@@ -108,11 +109,26 @@ int TimestampDiff::handler(int operation, String &data, Element *e,
...
     unsigned begin = 0;
-
-    if (data != "")
+    double perc = 0;
+    int opt = reinterpret_cast<intptr_t>(handler->user_data(Handler::f_read));
+
+    if (data != "") {
+        if (opt == TSD_PERC_HANDLER) {
+            int pos = data.find_left(' ');
+            if (pos == -1) pos = data.length();
```
gkatsikas: Re-check if pos should be equal to data.length() - 1 instead of data.length();

lib/dpdkdevice.cc
```diff
@@ -143,25 +150,27 @@ int DPDKDevice::alloc_pktmbufs()
...
                                 return rte_errno;
+                        }
                 }
         }
     } else {
         int i = 0;
         rte_mempool_walk(add_pool,(void*)&i);
         if (i == 0) {
-            click_chatter("Could not get pools from the primary DPDK process");
+            errh->error("Could not get pools from the primary DPDK process");
```
gkatsikas: Use directly return errh->error(..);

test/analysis/Replay-Timing-01.testie
```diff
@@ -0,0 +1,30 @@
...
+0.001 A
+0.002 B
+0.004 C
+0.004 D
+0.005 E
+
+%expect OUT
+!IPSummaryDump 1.3
+!data timestamp payload
+1000000000.000500{{[0-9]+}} "A"
```
gkatsikas: Are you sure about the stability of this testie?

PR #118, opened 2018-09-23
Batching support for VLAN elements

gkatsikas:


gkatsikas:
Indeed.. I might go a new patch soon

PR #120, opened 2018-09-24
Prevent accessing non-existing MAC header

gkatsikas:
This patch fixes an issue when attempting to set MAC
header on e.g., raw IP packets.

### DIFF
--- a/include/click/packet.hh
+++ b/include/click/packet.hh
@@ -2115,6 +2115,7 @@ Packet::set_dst_ip_anno(IPAddress a)
 inline void
 Packet::set_mac_header(const unsigned char *p)
 {
+    if (!p) return;
     assert(p >= buffer() && p <= end_buffer());
 #if CLICK_LINUXMODULE	/* Linux kernel module */
 # if LINUX_VERSION_CODE >= KERNEL_VERSION(2, 6, 24)


bcronje:
I don't think this is such a good idea. It would be much more beneficial to assert at this point than silently fail. If this is causing issues with another element like FromDump it should be fixed on that side in my opinion.

gkatsikas:
Yes I see your point.
My original intention was to somehow automatically detect the absence of MAC header (e.g., when FromDump generates raw IP packets) and let Click function while taking this into account.
I might follow one the following paths instead:
Either generate a warning in FromDump when raw IP packets are generated (simple and neat)
or traverse the downstream path and augment it with additional elements to auto-prevent faults; this sounds an overkill though.

tbarbette:
As said in the latest issue problem is also true for many IP elements. The warning should be avoided if the element is followed by a markmacheader, so that solution is not very clean either...

A better solution than per element traversal would be to consider annotations in general in a dynamic way. Only have metadata space for the annotations used by the elements in the configuration on one hand, then handle collisions better than the current horrible sharing of offsets in packet_anno.hh and then also detect when an element uses an annotation that is not set by a previous element in the path.
However this requires elements to declare all annotation they use, along with a write or read tag. I actually have a branch doing that but never merged it because the indirection leads to lower performance. But if we push the idea of static parameters, llvm backend etc, we may avoid the cost by generating the binary code for a specific metadata stack.

gkatsikas:
This solution requires a lot of work, which involves redesigning core parts of Click.
I am interested in taking parts of this redesign if a solid plan is devised.
Otherwise, I will stick to some temporary solution.

tbarbette:
Run-time checks, I very don't like (even more silently as bcronje said).

Maybe extends all Element::cast functions that do set the IP and MAC header to make them castable to MarkIPHeader and MarkMACHeader. Then we can ensure that at least an element before one using the MAC or IP header do cast in MarkIPHeader or MarkMACHeader. 
To avoid rewriting the visitor (though ElementCastVisitor can nearly be used as it), we could add a function to Element like

> ensure_upstream_cast("MarkMACHeader", errh);

 and we add that to all elements configure(...) that rely on the MAC/IP header being set.

PR #123, opened 2018-10-03
ip6address: fix parser array bounds issue/crash

pallas:
A malformed IP6Address, i.e. with more than 8 components, will cause the
memmove/memset at the end of parsing to fail.  This code is supposed to move
the post-:: bytes to the end of the address and zero out the bytes in the
center.  Instead, it moves some of the bytes the wrong way, potentially
writing prior to parts[] and then zeroing with a negative size.

Instead, make the parser stop when it hits 8 components and let parse figure
out that we did not consume the entire String.

Thanks: American Fuzzy Lop

Change-Id: I2966cb22fb5b44eb9e92dfc6a94e891cf5d02fa7


tbarbette:
Travis was in a bad shape, but the tests actually passed.

PR #127, opened 2018-10-26
Import from Metron

tbarbette:
A few things


gkatsikas:
Once you merge this one, could you also merge master into metron?

PR #131, opened 2018-11-06
Proper checksum calculation in FromDump with FORCE_LEN

gkatsikas:
This patch deals with FromDump when FORCE_LEN configuration
is active. In this case the checksum of each packet is entirely
recomputed to cover cases, such as anonymized traces.
The issue manifested itself when attempted to replay the new
anomymized KTH traces.

### DIFF
--- a/elements/userlevel/fromdump.cc
+++ b/elements/userlevel/fromdump.cc
@@ -519,15 +519,11 @@ FromDump::read_packet(ErrorHandler *errh)
             return false;
         }
 
-        // Keep the old length for incremental IP checksum calculation
-        uint16_t len_old = ip->ip_len;
-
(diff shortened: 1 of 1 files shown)


tbarbette:
Mhh this is a problem of the trace. If one pad a trace, he should know that a SetIPChecksum is needed. And also probably a SetTCPChecksum.  Or other protocols? So it seems a bit unclean to try to partially fix that here, no?

gkatsikas:
When FromDump is configured with FORCE_LEN = 0, Pad is not required at all as FromDump adjusts the length of the packet itself. However, I agree that checksum-specific elements should be used when needed. I will revoke the patch and close the PR.

PR #135, opened 2018-11-10
Force expensive packet reallocation

pallas:


tbarbette:
Why go through this PackerRef that is not really used? It justs complexifies a little the code... ?

pallas:
It's for debugging. A major issue is that many elements I've seen do not handle the expensive operations correctly. Since there were two refs, the old packet sticks around for awhile a send you don't get an immediate crash. This forces a reallocation and lights the issue up. The IPReassembler bugs were all pinpointed with this option.

~Derrick • iPhone

tbarbette:
This should probably be in some specific branch, no?

I thought about making also a research branch with more stuffs that should not go to mainline or never prove to increase performance in most cases. Maybe tool that will never be used at runtime such as this could go there ?

pallas:
It feels like the malloc debugging code, which is also just a default-disabled configure option. Putting it on a branch would make it harder to discover and my take is that it proved its value by highlighting decade-old bugs in core code.

~Derrick • iPhone

tbarbette:
Okay then. Thanks !

PR #136, opened 2018-11-10
ip/ipreassembler: fix positive header_delta push path

pallas:
Like the put earlier in this method, the push here can fail or reallocate
the SKB.  In the former case, we need to clean up (as above) and in both
cases, we need to update the linkage into q from it's predecessor.

Thanks: --enable-force-expensive + Valgrind

Signed-off-by: Derrick Pallas <pallas@meraki.com>


tbarbette:
Is my merge correct?

tbarbette:
With the other patch merged, can you rebase?

pallas:
Yes, will give it a shot today.

~Derrick • iPhone

PR #137, opened 2018-11-12
Adding 2 new conf to FastClick

lidall:
A pktgen script in order to generate traffic and receive them after passed through the DUT.
Besides, a single core NAT with DPDK enabled.


conf/nat/nat-DPDK.click
```diff
@@ -0,0 +1,91 @@
+/*
+ * This is a script based on Thomar NAT and with DPDK enabled. Also you can  
+ * just replace the FromDPDKDevice and ToDPDKDevice with FromDevice and ToDevice
+ * if you don't want to implement DPDK.
```
tbarbette: to use DPDK

conf/nat/nat-DPDK.click
```diff
@@ -0,0 +1,91 @@
+/*
+ * This is a script based on Thomar NAT and with DPDK enabled. Also you can  
+ * just replace the FromDPDKDevice and ToDPDKDevice with FromDevice and ToDevice
```
tbarbette: Queue -> ToDevice

conf/nat/nat-DPDK.click
```diff
@@ -0,0 +1,91 @@
+/*
```
tbarbette: Please change filename to be lowercase

Mr-Hongyi:
Actually, the pktgen-l3-latest.click is based on your script tester-l3.click. So you mean we just need to modify it and use the same filename to push?

tbarbette:
Yes please

Mr-Hongyi:
I think we are done and you can check it (:

tbarbette:
Thanks for the PR. Please look at my first correction commit though ;)

Things like NDESC, you can change it locally by overwriting ring_size but it's best to keep the default for examples.

PR #141, opened 2018-11-15
More packet memory stuff

pallas:


pallas:
As an aside, InfiniteSource is broken but the fastclick copy is different enough now from my copy that the patch is not clean.

tbarbette:
Should be fixed.This file was never modified in FastClick, I wonder why on your side the "q" got renamed in "p" :p 

I just wonder about the assert. Personally, I never disable them, so it will cost a little. Is it really necessary? If we don't add "null" to the hashtable, we'll never get null, right?

tbarbette:
I'll let you rebase, and remove what's unneeded. And for the assert?

bcronje:
> I just wonder about the assert. Personally, I never disable them, so it will cost a little. Is it really necessary? If we don't add "null" to the hashtable, we'll never get null, right?

`i.get()` here refers to the hashtable element, not the value. So I think it could be that OOM can cause this assert to fail. However we also dont disable asserts and operator[] could possibly be used very frequently so I'm not sure this is a good idea. How about using `click_hash_assert` as defined in hashcontainer.hh where CLICK_DEBUG_HASHMAP should be enable for the assert to take affect?

PR #146, opened 2018-11-23
Support for DPDK 18.11

gkatsikas:
A problem still occurs when line 358 in dpdk.mk is executed.
For some reason the make process attempts to copy a file (tm.c)
that is non-existent anymore in DPDK 18.11.
I guess this is somewhere cached and needs to be flushed before
cp and gcc are invoked.


tbarbette:
I"ll look at this after the deadline. Too much breaking changes already in...

gkatsikas:
I took some time in the weekend to fix the compatibility issues and make travis happier.
The issue with the file copy persists for DPDK 18.11 though.

gkatsikas:
The only pending issue here is a bunch of new structs in testpmd.h which get multiply defined. We need to follow the same approach as with the other structs, using extern and sed :(

gkatsikas:
To name those structs, here we are:

mplsoudp_decap_conf
mplsoudp_encap_conf
mplsogre_decap_conf
mplsogre_encap_conf
l2_decap_conf
l2_encap_conf

PR #148, opened 2018-11-23
More batch-compatibility

piotrjurkiewicz:
Add batching support for various elements.


tbarbette:
The failures of travis seem legit. Can you take a look? https://travis-ci.org/tbarbette/fastclick/builds/458877565?utm_source=github_status&utm_medium=notification

piotrjurkiewicz:
I updated the PaintTee commit to use `BATCH_CREATE_*` macros instead `PacketBatch::append_packet` in loop.

Rate shaping elements are problematic with batching. In order to maintain exactly the same results as in case of non-batching operation, they would have to apply per-packet granularity (refill token bucket after each packet in a batch and split that batch/pull from queue on per-packet basis), what would take away any performance gains provided by batching operation. These differences are only meaningful when token bucket capacity is low comparing to batch size.

tbarbette:
I merged this anyway. What I did for those element like RRSwitch is to add a "PER_BATCH" argument that would allow considering a per-batch round robin, instead of splitting the batch.

PR #149, opened 2018-11-23
Fix FromNetmapDevice timestamps and ToNetmapDevice arguments parsing

piotrjurkiewicz:
Also change indentation in FromNetmapDevice to 4 spaces. The file contained mixed portions of tabs, spaces and tabs-with-spaces indentation. I find such an indentation mess very problematic, especially in files with convoluted logic like this. Maybe its high time to introduce consistent space-only indentation across the whole FastClick codebase?


tbarbette:
Ok yes I remember now. The ts for the first packet makes sense because anyway the whole batch is received at once. However Click would have to be modified to support the idea of per-batch annotation.

PR #150, opened 2018-11-23
IPOutputCombo: stop click_chatter on PaintTee

piotrjurkiewicz:


tbarbette:
Yes, this was to display clearly the error path, but the user should handle that.

tbarbette:
Applied, thanks !

PR #151, opened 2018-11-23
Fix GROW bug in StoreData and introduce data MASK

piotrjurkiewicz:


tbarbette:
Great ! Thanks ! And sorry for the delay ;)

PR #152, opened 2018-11-23
The new CounterFile element, based on CounterAtomic

piotrjurkiewicz:
CounterFile element writes packet and byte counts to a mmapped-file
providing external processes a fast and low-overhead way to read
counter values.


tbarbette:
Could you describe how the thread safeness works in the documentation?

So elements will use atomic operation over the memory that is mapped. How will that work with the underlying file?

PR #155, opened 2018-12-03
String split returns empty vector when no input

gkatsikas:

### DIFF
--- a/lib/string.cc
+++ b/lib/string.cc
@@ -611,13 +611,11 @@ String::find_left(char c, int start) const
     return -1;
 }
 
-/** @brief Search for a character in a string.
- * @param c character to search for
- * @param start initial search position
+/** @brief Split a string using an input character as a delimiter.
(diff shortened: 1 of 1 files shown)


bcronje:
Could you maybe update String::split documentation as well, as it currently has find_left documentation?

tbarbette:
Thanks :p I was probably not into documentation when I pushed that ^^

tbarbette:
Oh but wait, if I just asks to split "hello" with the "," vector, what will I get?
The normal behavior of split in other languages is to pass a vector with only "hello". You did not change that right?

gkatsikas:
No the patch does not work like that :p
In this case it will return an empty vector.

The problem is when the input string is empty.
In this case the old code returns a vector with 1 element, which is the empty string.
Ideally, we want it to return an empty vector, don't we?

I could modify the patch to return an empty vector if the string is empty before the loop starts. Then the rest of the method will be the same as before this patch.

PR #157, opened 2018-12-12
Batching in FromDevice.u and ToDevice.u elements

piotrjurkiewicz:
I have done it mostly to avoid batch mode warnings on startup and did not expect performance gains. Nevertheless, I observe 8-10% higher throughput with batching.


tbarbette:
Thanks !

PR #161, opened 2019-01-02
Conditional kill of packet in ensure DPDK buffer

gkatsikas:
This is a way to circumvent memory corruption when trying to ensure the buffer space of large DPDK frames

### DIFF
--- a/elements/userlevel/ensuredpdkbuffer.cc
+++ b/elements/userlevel/ensuredpdkbuffer.cc
@@ -65,14 +65,14 @@ EnsureDPDKBuffer::smaction(Packet* p) {
     } else if (_noalloc) {
         p->kill();
         if (_warn_count++ < 5)
-            click_chatter("%p{element} : Not a DPDK packet",this);
+            click_chatter("%p{element}: Not a DPDK packet", this);
         return 0;
     } else {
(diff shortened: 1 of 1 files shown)


elements/userlevel/ensuredpdkbuffer.cc
```diff
@@ -88,7 +88,10 @@ EnsureDPDKBuffer::smaction(Packet* p) {
             p->kill();
             return q;
         } else {
-            p->kill();
+            click_chatter("[%s] Could not ensure buffer for packet with length %d bytes. Dropped!", name().c_str(), p->length());
+            if (p->length() <= (DPDKDevice::MBUF_DATA_SIZE - RTE_PKTMBUF_HEADROOM)) {
```
tbarbette: I don't understand this? Where the this length computation comes from? What should the user assume?

gkatsikas: When input frames exceed this length FastClick fails to release the allocated memory (i.e., p->kill() causes seg. fault when it eventually calls the destructor of such a packet). The problem is deeper, this condition just gives a hint on what part of memory allocation might be problematic

tbarbette:
Ok let's do that for now, but it would be great to solve the underlying problem...

PR #166, opened 2019-02-28
BandwithRatedUnqueue: Allow EXTRA_LENGTH and use link rate

tbarbette:


elements/standard/bwratedunqueue.cc
```diff
@@ -19,13 +19,28 @@
...
+    if (Args(this, errh).bind(conf)
+        .read_or_set("EXTRA_LENGTH", _use_extra_length, false)
+        .read_or_set("LINK_RATE", _link_rate, false)
+        .consume() < 0)
+        return -1;
+    return RatedUnqueue::configure(conf, errh);
+}
+
+#define LENGTHOF(p) \
+    ((_link_rate?8:1)*(p->length() + _use_extra_length? EXTRA_LENGTH_ANNO(p) : 0) + (_link_rate? 8 * 24:0))
```
gkatsikas: I suppose that the purpose here is to convert into bits/sec over Ethernet when link_rate=true.

(_link_rate?8:1) converts a packet's length (in bytes) to bits.
(_link_rate? 8 * 24:0) is the sum of 1. Inter Frame Gap (12 bytes) + 2. MAC Preamble (8 bytes) + 3. CRC (4 bytes) = 24 bytes. We multiply by 8 to convert to bits and we are set.

Looks good to me :p
I just wanted to verify it analytically :)

elements/standard/bwratedunqueue.cc
```diff
@@ -43,7 +58,7 @@ BandwidthRatedUnqueue::run_task(Task *)
             PacketBatch* batch = input(0).pull_batch(_burst);
             if (batch) {
                 int c = 0;
-                FOR_EACH_PACKET(batch, p) c += p->length();
+                FOR_EACH_PACKET(batch, p) c += LENGTHOF(p);
```
gkatsikas: isn't there a pull_batch here?

elements/standard/bwratedunqueue.cc
tbarbette: Yes you're right. It's not like if I messed it up in the past :p

PR #170, opened 2019-03-20
Enabling/disabling VLAN offload features in FromDPDKDevice

reuvenperetz:
Adding support of VLAN offload features such as filtering of received VLAN packets and VLAN header stripping by hardware in received VLAN tagged packets.
This commit changes the default behavior and enables turning on/off these offloads, by adding two new boolean arguments to FromDPDKDevice element: VLAN_STRIP and VLAN_FILTER.

### DIFF
--- a/elements/userlevel/queuedevice.cc
+++ b/elements/userlevel/queuedevice.cc
@@ -102,8 +102,12 @@ int RXQueueDevice::parse(Vector<String> &conf, ErrorHandler *errh) {
     QueueDevice::parse(conf, errh);
 
 	_promisc = true;
+	_vlan_filter = true;
+	_vlan_strip = true;
     if (Args(this, errh).bind(conf)
 	.read_p("PROMISC", _promisc)
+	.read_p("VLAN_FILTER", _vlan_filter)
+	.read_p("VLAN_STRIP", _vlan_strip)
         .consume() < 0)
         return -1;
 
--- a/elements/userlevel/fromdpdkdevice.cc
+++ b/elements/userlevel/fromdpdkdevice.cc
@@ -125,7 +125,7 @@ int FromDPDKDevice::initialize(ErrorHandler *errh)
     if (ret != 0) return ret;
 
     for (unsigned i = (unsigned)firstqueue; i <= (unsigned)lastqueue; i++) {
-        ret = _dev->add_rx_queue(i , _promisc, ndesc, errh);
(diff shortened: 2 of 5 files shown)


elements/userlevel/queuedevice.cc
```diff
@@ -102,8 +102,12 @@ int RXQueueDevice::parse(Vector<String> &conf, ErrorHandler *errh) {
     QueueDevice::parse(conf, errh);
 
 	_promisc = true;
+	_vlan_filter = true;
```
tbarbette: Sorry I missed it up to now, but I would expect the default behavior to strip VLAN but not to filter VLANs. What would be filter filtering on by default? Can we set an index somehow?

reuvenperetz: You're right, my bad.
What does "set an index" mean?

tbarbette: Well the filtering feature allows to filter some VLAN ID I guess, right? 
I see the function rte_eth_dev_vlan_filter takes an ID. is The VLAN_FILTER parameter you added even useful without a way to set the ID being filtered?

reuvenperetz: Yeah, I see what you mean... Well, I used rte_eth_dev_set_vlan_offload to set it in general (because my click configuration needs to get the entire range < 4096, so VLAN_FILTER is useful for me), but I think you're right and it would be wiser to fix it so it can be configured by VLAN id (and I'm guessing it's relevant for stripping as well)...I'll see how can I modify and test it.
Thanks a lot for your feedback!

tbarbette: What is the default behavior when no ID is specifically set? If it accepts all VLAN, we can merge this as it and then you can extend. But I don't see where this is defined...

reuvenperetz: Yes, it accepts all VLANs. I don't know if it counts as documentation but rte_eth_dev_set_vlan_offload [description](https://doc.dpdk.org/api/rte__ethdev_8h.html#a81a17c2c85724d5222ab137188599a28) is the best I could find... But maybe it's best to rewrite it with the support of filtering/stripping specific vlan ids.

tbarbette:
Let's take this step in and provide a way to select the vlan ids later.

tbarbette:
Thanks !

PR #171, opened 2019-03-30
Timed source Segmentation Fault

ekline2:
Timed source seg faults when Click is built with DPDK (appears to work without DPDK).  It appears the cause is attempting to create the "source packet" before the packet pool is ready.  Moving packet creation from Configure to Initialize (similar to other sources) solves the issue.


tbarbette:
Yes, makes sense. Thanks!

PR #177, opened 2019-04-22
VLAN offloading init moved

gkatsikas:
VLAN offloading init moved after Rx queues' init
as it was causing segmentation fault, due to NULL
rxq object in mlx5_vlan_strip_queue_set().


tbarbette:
Thanks !

PR #179, opened 2019-05-01
TokenBucket: use 64bit if supported

tbarbette:
Allows all elements such as ratedsource, ratedsplitter, ratedunqueue,
and bandwidthratedunqueue to use > 4G rate.

tokenbucket64 could have lass copy pasting from tokenbucket, but I don't have time for this... I also changed the way it works a little. 32bit token buckets use overflow value as MAX. The 32bit value must therfore be scaled.
While with 64bit, I prefer to keep a reference to the max, and have easier and faster calculation.

### DIFF
--- a/elements/standard/ratedsource.cc
+++ b/elements/standard/ratedsource.cc
@@ -50,7 +50,7 @@ RatedSource::configure(Vector<String> &conf, ErrorHandler *errh)
     String data =
       "Random bullshit in a packet, at least 64 bytes long. Well, now it is.";
     unsigned rate = 10;
-    unsigned bandwidth = 0;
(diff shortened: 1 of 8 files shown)


tbarbette:
It seems travis does not understand uint64_t... To be continued.

pallas:
This should be done with templates and feature checks, right?

tbarbette:
> This should be done with templates and feature checks, right?

Well in the end the 32bit and 64bit version do not work the same way as we do not need overflow checks and this max value trick. But some factorization could happen, yes... I tried the templating approach first and abandoned...

tbarbette:
Done in #395

PR #180, opened 2019-05-05
fix octeon dpdk make rules for v18.05

rcgoodfellow:
dpdk v18.05 includes the octeontx family of libraries which get picked up by fastclick, but `librte_common_octeontx` is specifically gated by dpdk versions >= v18.09 which causes linker errors. this commit lowers the version gate to v18.05 to allow fastclick to link without error.


tbarbette:
Thanks !

PR #182, opened 2019-05-08
Series of patches for 100G Trace capture

tbarbette:
This serie adds support for timestamp hardware offloading (when DPDK is patched, hopefully, they will accept my patch upstream). 

ConnectX5 NIC can write the device's internal Clock value (much like the TSC of x86 CPUs) to the packet timestamp field. From there TSCClock can be used to convert the clock ticks to real time.

Example : 
```
fd :: FromDPDKDevice($port, BURST 32, PROMISC true, VERBOSE 99, PAUSE none, 
     MAXTHREADS 4, MTU $mtu, SCALE parallel, NUMA true, NUMA_NODE 1, TIMESTAMP true, 
     NDESC 4096); 
        -> tc :: TSCClock(INSTALL false, SOURCE fd, READY_CALL sl.run)  
        -> ... // ToDump etc

sl :: Script(TYPE PASSIVE,                
        wait 1s,                            
        print "Now recording!"
        write sact.switch 0 //To avoid writing packets with random timestamp, we have a Switch element to drop packet until the TSCClock has converged
);   
```

Then a few fixes and features encountered when doing high speed trace capture.

### DIFF
--- a/elements/analysis/timestampaccum.hh
+++ b/elements/analysis/timestampaccum.hh
@@ -22,7 +22,13 @@ timestamp. Keeps track of the total elapsed time accumulated over all packets.
 Returns the number of packets that have passed.
 
 =h time read-only
-Returns the accumulated timestamp difference for all passing packets.
+Returns the accumulated timestamp difference for all passing packets in seconds.
+
+=h min read-only
+Returns the minimal timestamp difference across all passing packets in seconds.
+
+=h max read-only
+Returns the maximal timestamp difference across all passing packets in seconds.
 
 =h average_time read-only
 Returns the average timestamp difference over all passing packets.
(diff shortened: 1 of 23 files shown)


elements/analysis/timestampaccum.hh
```diff
@@ -22,7 +22,13 @@ timestamp. Keeps track of the total elapsed time accumulated over all packets.
 Returns the number of packets that have passed.
 
 =h time read-only
-Returns the accumulated timestamp difference for all passing packets.
+Returns the accumulated timestamp difference for all passing packets in seconds.
+
+=h min read-only
```
tbarbette: Poke @aliireza . You may use those

PR #186, opened 2019-07-30
Preliminary support for DPDK 19.08

gkatsikas:
This patch supports up to DPDK 19.08-rc2.
More work migth be needed towards 19.08.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache:
     - dpdk-16.04
     - dpdk-17.08
     - dpdk-18.08
+    - dpdk-19.08
     - tcpdump-4.7.4
     - netmap-11.1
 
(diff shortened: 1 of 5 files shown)


tbarbette:
I pushed the test for 19.08 but it will not work as the package is not yet available, even for rc ( see http://fast.dpdk.org/rel/ ) should we wait or already merge?

gkatsikas:
I suggest to wait until the very last rc version, as things might change.

gkatsikas:
FastClick compiles against DPDK 19.08-rc4 without any changes.
We are approaching the stable release.

gkatsikas:
19.08 is upstreamed, merge looks safe

tbarbette:
Ok, just re-started the CI, then one of us can merge when it goes through.

tbarbette:
A few fixes due to new merged-in code and change in DPDK target names.
Thanks !

PR #194, opened 2019-08-07
Declare IP_OFFMASK macro only in kernel

gkatsikas:
The IP_OFFMASK macro is present in <netinet/ip.h>
thus user-level Click must refer to this macro
instead of redefining it.


tbarbette:
Thanks !

PR #198, opened 2019-10-11
IPPair unnecessarily re-defined in LinkTable

gkatsikas:

### DIFF
--- a/elements/wifi/linktable.hh
+++ b/elements/wifi/linktable.hh
@@ -20,39 +20,6 @@ CLICK_DECLS
  * =a ARPTable
  *
  */
-class IPPair {
-  public:
-
-    IPAddress _to;
-    IPAddress _from;
-
-    IPPair()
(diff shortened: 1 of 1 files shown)


tbarbette:
And it does not need any include on top of it?

gkatsikas:
Apparently no extra includes are needed :)
Let's merge

tbarbette:
And you compiled with wifi enabled ?  Because redefinition prevents compilation so it would not have worked before.

Le 11 oct. 2019 23:01, Georgios Katsikas <notifications@github.com> a écrit :

Apparently no extra includes are needed :)
Let's merge

-

gkatsikas:
Yes, initially I compiled with wifi enabled and got the redefinition error.
After the commit, it works fine.
If you disable wifi, linktable is not compiled at all, so no errors either.

tbarbette:
Okay then !  Thanks !

PR #200, opened 2019-10-15
LRO, Rx jumbo frames, and VLAN extend offloading

gkatsikas:
To allow Rx jumbo frames offloading it is necessary
to set max_rx_pkt_len, which was not set by FastClick.
Also, fixed the initial VLAN filter and strip configuration
status (which was mistakenly set to true).
Messages are added to notify users about the status of
the offloading features when turned on.

### DIFF
--- a/elements/userlevel/fromdpdkdevice.cc
+++ b/elements/userlevel/fromdpdkdevice.cc
@@ -120,7 +120,7 @@ int FromDPDKDevice::configure(Vector<String> &conf, ErrorHandler *errh)
         _dev->set_rx_offload(DEV_RX_OFFLOAD_TIMESTAMP);
         _set_timestamp = true;
 #else
-        errh->error("HW Timestamping is not supported before DPDK 18.02");
+        errh->error("Hardware timestamping is not supported before DPDK 18.02");
 #endif
     }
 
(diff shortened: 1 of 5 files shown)


gkatsikas:
I will fix compatibility issues asap

tbarbette:
Seems good after rebasing. I also fixed all spaces in queuedevice.cc

tbarbette:
I forced-push a rebase to verify it's compiling now

gkatsikas:
Hopefully my last patch will solve the issue with 17.08

gkatsikas:
> No it is not. This offloading feature has been around since the time DPDK was still under the Intel umbrella.
> I didn't know what was the first version supported by FastClick (probably 2.2 or 2.0), so I put it this way

Finally, it makes no sense to keep such an old compatibility with DPDK 1.7.0, because to set this offload you need to access different data structures which changed several times until 19.08.
In order to keep the code readable, I simply demand DPDk 17.11 or higher for this feature (as happens with most offloading features present in this library).

gkatsikas:
I rebased the commits into one to allow easier merge

tbarbette:
Oh yeah, sorry. I forgot this one. Thanks !

PR #201, opened 2019-10-15
Fixed compatibility issue with DPDK 19.08

gkatsikas:


tbarbette:
See fix from #202. Also is it intended you pushed on metron?

gkatsikas:
> See fix from #202. Also is it intended you pushed on metron?

Yes, I intentionally pushed to metron :p
Could you merge #202 to metron?

PR #203, opened 2019-10-15
Address issues with DPDK ether_addr

gkatsikas:


gkatsikas:
The problem appears to be travis.
The "problematic" testie is ./test/IPRewriter/TCPRewriter-01-StateMigration.testie, which fails with both gcc and clang.
However, it is only reported as problematic in the case of clang.

tbarbette:
It's an illegal instruction. I guess it's one of the only test to actually use DPDK. And DPDK compiles by default with fancy new instructions not supported by travis.

tbarbette:
I think we should add a few lines of sed in the travis script to deactivate fancy things, or force DPDK to use a specific lower level architecture such as core2 (there is some DPDK options for that). Do you want me to look for it?

gkatsikas:
It would be nice of you if you could take a look at that!

PR #207, opened 2019-10-23
Batching support for AggregateLength element

gkatsikas:


tbarbette:
I think SimpleElement<AggregateLength> would have been enough. But thanks !

PR #208, opened 2019-10-23
Additional statistics for AverageBatchCounter

gkatsikas:
These statistics are related to the frame lengths being
observed either since the last tick or since the initialization
of this element. An input parameter defines whether these extra
statistics will be provided or not. Also, the element's documentation
is improved.


tbarbette:
Thanks !

PR #212, opened 2019-11-08
IPFilter caching and style changes

gkatsikas:
To see the actual changes introduced by caching, please check the first commit only.
The efficiency of the caching could be further improved if we use the RSS hash directly from the hardware, but this is left as future work :p

### DIFF
--- a/elements/ip/ipfilter.hh
+++ b/elements/ip/ipfilter.hh
@@ -2,7 +2,6 @@
 #define CLICK_IPFILTER_HH
 #include "elements/standard/classification.hh"
 #include <click/batchelement.hh>
-#include <click/hashmap.hh>
 #include <click/ipflowid.hh>
 #include <click/error.hh>
 CLICK_DECLS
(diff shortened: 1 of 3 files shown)


elements/ip/ipfilter.hh
```diff
@@ -323,28 +372,55 @@ inline bool
...
+        if (caching) {
+            (*_cache_misses_nb.findp(name()))++;
+        }
+        // common case never checks packet length
+        return length_checked_match(zprog, p, packet_length);
+    }
+
+    // Caching enabled
+    if (caching) {
+        if (_last_flow_id[name()]) {
```
tbarbette: Er... multiple hashtable search in the fast path? :p Why not a per-element cache?

gkatsikas: Good catch, thanks! I was already fixing the problem as I realized myself how stupid it was :p

gkatsikas: Do not merge yet, as I want to test the performance of caching with my new testie.
I will give you the green flag once I finish.
Of course, feel free to propose more improvements ;)

gkatsikas: Performance looks good in general. With a small number of rules, there is a very light impact of caching on the latency, but with a large number of rules the performance gains can be substantial. Green flag from me :p

tbarbette:
Yep seems good! Thanks !

Issue #111, opened 2018-08-31
Which model for hardware offloading?


tbarbette:
There are basically 2 ways :
- Implement everything in the From/ToDPDKDevice elements. But that makes I/O elements very complex, with a lot of parameters to turn offloadings on and off, and code to support backup solution when hardware does not support the feature. 
- Add feature-element like SetTimestamp to set the timestamp, SetTCPChecksum, .... And somehow if the timestamp is provided by the hardware, this element would be "deactivated" and the hardware one used.
The second one is more "Click-like" because the logic is specified by elements. But how to make the element "discuss" to establish that the feature is supported by the upfront hardware?

tbarbette:
What to do first : 
- Hardware timestamp + software backup in FromDPDKDevice
- Support for RX/TX checksum offloading

tbarbette:
Poke @gkatsikas  I need to solve this to implement the Metron monitoring (for the timestamp). What's your take on this? Others are welcome to share their thoughts of course.

cffs:
The Click-like solution seems more elegant to me. Maybe we could enrich `DPDKDevice` with more functionality so that offloading features can be queried/set from various elements at configure/initialization time.

tbarbette:
But then how does, eg, SetTCPChecksum finds the downstream transmitter(s).
Eg.
```
SetTCPChecksum -> Something -> classifying :: Classifier (...) ;
classifying[0] -> ToDPDKDevice(1);
classifying[1] ->  Something -> ToDPDKDevice(2)
```
One idea would be to look downstreamfrom SetTCPChecksum for elements that can cast themselves as "SetTCPChecksum", and then ask the casted one to enable the functionnality.

cffs:
Well, I had only thought about the receiving side up to that point. There, it is feasible to check if a timestamp/checksum is already there, and compute it otherwise. On element output, the timestamp/checksum is guaranteed to be there.

Enabling offloading on the input device is orthogonal. It could either be automatic (*e.g.* `SetTCPChecksum` would look at all its parent input devices, and try to enable offloading), or manual (by adding parameters directly to `FromDPDKDevice`).

On the transmission side, it is not as straightforward. What if some packets go to a device that support offloading, and some go to a device that does not support it? Moreover, it feels wrong to have a `SetTCPChecksum` element after which TCP checksum is still invalid, waiting for some output device to change it further down the path.

[...]

PR #214, opened 2019-11-18
Metron extensions and bug fixes

gkatsikas:
New configuration parameters FLOW_DIR_ISOLATE in FromDPDKDevice.
If set != 0, this parameter guarantees that all ingress traffic
comes from defined flow rules only (current and future), thus
rendering PMD more deterministic. It can be configured only if the
device is in MODE flow_dir. Handlers are added in FromDPDKDevice to
read the current value of this parameter and reset it at will.

Fixed bug in FromDPDKDevice queue_count and queue_bytes handlers.

Impoved the documentation of FromDPDKDevice and ToDPDKDevice elements
by adding description for all configuration parameters and handlers.
Some parameters mentioned in FromDPDKDevice element belong to ToDPDKDevice,
therefore they are moved accordingly.

FlowDirector supports flow isolation through the abovementioned
FromDPDKDevice handler. Also, handler 'rules_list_with_hits' is
added to display only the rules that exhibit packet hits.
The output of handler 'rules_aggr_stats' is nicely formatted to
display per queue load balancing statistics as well as aggregate
statistics.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ cache:
   directories:
     - dpdk-2.2.0
     - dpdk-16.04
+    - dpdk-17.05
     - dpdk-17.08
     - dpdk-17.11
     - dpdk-18.02
--- a/elements/userlevel/fromdpdkdevice.cc
+++ b/elements/userlevel/fromdpdkdevice.cc
@@ -73,8 +73,8 @@ int FromDPDKDevice::configure(Vector<String> &conf, ErrorHandler *errh)
     int max_rss = 0;
     bool has_rss = false;
 #if RTE_VERSION >= RTE_VERSION_NUM(17,5,0,0)
-    String fd_rules_filename;
-    bool fd_isolate = false;
+    String flow_rules_filename;
+    bool flow_isolate = false;
 #endif
     if (Args(this, errh).bind(conf)
         .read_mp("PORT", dev)
@@ -90,8 +90,8 @@ int FromDPDKDevice::configure(Vector<String> &conf, ErrorHandler *errh)
         .read("MTU", mtu).read_status(has_mtu)
         .read("MODE", mode)
     #if RTE_VERSION >= RTE_VERSION_NUM(17,5,0,0)
-        .read("FLOW_DIR_RULES_FILE", fd_rules_filename)
-        .read("FLOW_DIR_ISOLATE", fd_isolate)
+        .read("FLOW_RULES_FILE", flow_rules_filename)
+        .read("FLOW_ISOLATE", flow_isolate)
     #endif
         .read("VF_POOLS", num_pools)
         .read_all("VF_VLAN", vf_vlan)
@@ -118,14 +118,14 @@ int FromDPDKDevice::configure(Vector<String> &conf, ErrorHandler *errh)
 
     int r;
     if (n_queues == -1) {
-    if (firstqueue == -1) {
-        firstqueue = 0;
-        // With DPDK we'll take as many queues as available threads
-        r = configure_rx(numa_node, 1, maxqueues, errh);
-    } else {
-        // If a queue number is set, user probably wants only one queue
-        r = configure_rx(numa_node, 1, 1, errh);
-    }
+        if (firstqueue == -1) {
+            firstqueue = 0;
+            // With DPDK we'll take as many queues as available threads
+            r = configure_rx(numa_node, 1, maxqueues, errh);
+        } else {
+            // If a queue number is set, user probably wants only one queue
+            r = configure_rx(numa_node, 1, 1, errh);
+        }
     } else {
         if (firstqueue == -1)
             firstqueue = 0;
@@ -158,11 +158,7 @@ int FromDPDKDevice::configure(Vector<String> &conf, ErrorHandler *errh)
         _dev->set_init_rss_max(max_rss);
 
 #if RTE_VERSION >= RTE_VERSION_NUM(17,5,0,0)
-    if (fd_isolate && (mode != FlowDirector::FLOW_DIR_MODE)) {
-        return errh->error("Flow Director (port %s): FLOW_DIR_ISOLATE requires MODE flow_dir", dev.c_str());
-    }
-
-    if ((mode == FlowDirector::FLOW_DIR_MODE) && (fd_rules_filename.empty())) {
+    if ((mode == FlowDirector::FLOW_DIR_MODE) && (flow_rules_filename.empty())) {
         errh->warning(
             "Flow Director (port %s): FLOW_DIR_RULES_FILE is not set, "
             "hence this NIC can only be configured by the handlers",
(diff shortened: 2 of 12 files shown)


tbarbette:
The option in RSS++'s branch is named ISOLATE. As it's more related to rte_flow than Flow director, it's probably best to keep that name (and for compatibility of scripts).
I think it would also be a good time to push a solution to remove the testpmd build script. That's the only thing preventing merging... Maybe on DPDK side?

gkatsikas:
Actually, there is a naming confusion here and it is good that you mentioned it.
Flow Director is an Intel feature related to flow steering introduced long time ago.
This feature is present in other NICs too (e.g., Mellanox) with a different name.
DPDK tried to homogenize flow steering across all vendors that support DPDK under a common umbrella called DPDK Flow API, which uses rte_flow as a fundamental data structure for flow steering. So to me, rte_flow and Flow Director are the same thing in DPDK terms.

The actual purpose of the lib/flow_director.cc and include/click/flow_director.hh library was to incorporate the DPDK Flow API into FastClick, but I used a bad name (FlowDirector instead of e.g., DPDKFlow). So, my first task is to refactor the library to avoid confusion and modify the MODE argument (instead of flow_dir it should be flow_api or simply flow).

You are right, this annoying dependency on testpmd which requires to copy and modify parts of testpmd locally should be removed, but I don't know if the solution will satisfy DPDK people. It will also take some time to implement a workaround that keeps both DPDK and FastClick satisfied.

Regarding the ISOLATE parameter, I believe it is important to add a prefix that associates this parameters with the MODE of the FromDPDKDevice element. This way the user understands that the isolation is at the level of the Flow Steering mechanism in the NIC and not some other type of isolation.

Finally, an even more generic solution would be to further generalize this library by being able to support both DPDK (rte_flow) and regular Linux-based (e.g., through ethtool) flow steering. So in the future both FromDPDKDevice and regular FromDevice elements could exploit this library. We should probably organize this information as a 'Project', where we could write down the tasks for future development of this library along with their priorities. Priority 1 is to get rid of the dependency and perform a simple refactoring of the library to avoid the confusion with Intel's Flow Director.

tbarbette:
In the RSS++ branch, both FromDevice and FromDPDKDevice define common functions of a parent EthernetDevice structure to do some common functions, like setting the indirection table, querying its size etc. But for flows does it really make sense? There are very different limitations... I would stick to DPDK-only.

gkatsikas:
FLOW_ISOLATE is ok.
I see what you mean, RSS can also be modified through rules so I should decouple FLOW_ISOLATE from the MODE.
For now, let's keep Flow API only under DPDK, but it would be nice to detach all offloading features from From/ToDPDKDevice elements and incorporate them into a common library/element as discussed in #111.

Regarding the refactoring of lib/flow_director.cc, this should happen after we remove the testpmd dependency, merge RSS++ with Metron and then both with master.

gkatsikas:
If we want to completely get rid of testpmd, there are two solutions:

1. abstract the flow parser from testpmd and make it a new library e.g., librte_flow_parser or part of librte_cmdline.
I sent a patch for DPDK 18.02 and although some people appreciated this separation, some others expressed their objections about moving the parser out of testpmd, so it did not go through in the end.
They were positive only if the parser becomes super generic. which also requires a lot of work.
I can forward their responses to you, just to get an idea of what they thing.
2. to implement our own flow rule parser and directly call the rte_flow functions, without involving the corresponding testpmd wrappers (port_flow_* functions).
This requires quite a lot of work and results in (i) code duplication and (ii) continuous maintenance and updates when new fields are added to the DPDK parser.

What do you propose?

gkatsikas:
Alternative to the above two options for now is to send a minimal patch to DPDK 19.11 which will result in getting rid of all the sed commands in our mk file, apart from those 2 which strip off the printf commands (e.g., Flow rule create/destroy).
The only commands that will remain in our mk file will simply copy testpmd locally and strip off the main function:
	mkdir -p $(PARSE_PATH)
	mkdir -p test-pmd
	# Copy testpmd.c of the target DPDK version
	cp -u $(RTE_SDK)/app/test-pmd/testpmd.c $(PARSE_PATH)
	cp -u $(RTE_SDK)/app/test-pmd/testpmd.h $(PARSE_PATH)
	cp -u $(RTE_SDK)/app/test-pmd/config.c $(PARSE_PATH)
	# Strip the main function off to prevent complilation errors, while linking with Click
	sed -i '/main(int/Q' $(PARSE_PATH)/testpmd.c
	head -n -1 $(PARSE_PATH)/testpmd.c > $(PARSE_PATH)/testpmd_t.c;
	mv $(PARSE_PATH)/testpmd_t.c $(PARSE_PATH)/testpmd.c
	sed -i '/printf("Flow rule #\%u created\\n", pf->id);/d' $(PARSE_PATH)/config.c
	sed -i '/printf("Flow rule #\%u destroyed\\n", pf->id);/d' $(PARSE_PATH)/config.c

This is a good way forward but does not entirely solve our problem

tbarbette:
Let's dig up on (1) a little.  Maybe instead of separation, we can simply just ask for what we do with sed.

gkatsikas:
If I haven't missed anything, this patch along with the upcoming v19.11 substantially improves the build process of the metron branch. The number of seds is greatly minimized and we could bring back the parallel compilation flag. At or after 19.11, the only seds performed are (i) to remove the main function from testpmd.c and strip off (ii) Flow rule created (iii) Flow rule deleted commands from config.c.

I believe travis tests must be increased to cover at least 17.05 and 19.11 builds

tbarbette:
You can add it in travis.yaml.
Then for the NOTPARALLEL, this seems to work in dpdk.mk:
ifneq ($(shell [ -n "$(RTE_VER_YEAR)" ] && ( ( [ "$(RTE_VER_YEAR)" -eq 19 ] && [ "$(RTE_VER_MONTH)" -ge 11 ] ) || [ $    (RTE_VER_YEAR) -ge 20 ] ) && echo true),true)
.NOTPARALLEL: ${PARSE_PATH}.sentinel
endif

gkatsikas:
Yes, I know how to do it, I was just asking your opinion :p
Maybe you did not want them

tbarbette:
I was not sure that the .NOTPARALLEL would not always be taken in even in between a conditional close so I tried locally.
Why the duplication with port_flow_isolate in lib now that it is a "pure" DPDKDevice function?
Else I'm okay with it.

gkatsikas:
It is not a duplication, flow isolation is achieved only through Flow Director library, because it is also part of the rte_flow_* functions

tbarbette:
Maybe for merging to master, we can have an --enable-metron that will lead to "notparallel" with old DPDK versions. And have metron built by default starting at 19.11.

gkatsikas:
Sounds like a good solution. Bare in mind that we should not remove the metron branch because it is referenced by the NSDI paper (reference [5]). So, once we merge, the metron branch will just be a mirror of master

tbarbette:
Sure. Anyway the merge should be manual, in multiple packages. There are things to refactor such as the thread reconfiguration message.

gkatsikas:
I suggest to start gradually by first merging RSS++ into metron and test that both work (e.g., reproducing the CoNEXT and NSDI/TOCS experiments).
Then, we should go to master.

tbarbette:
OK. So much fun to go :p
I'll merge the flow facilities of RSS++ in a parallel track, as discussed before.

PR #215, opened 2019-11-24
FastClick v19.11

tbarbette:
Hi all,

After solving a huge source of cache miss in DPDK TX, I thought it would be nice to version a bit FastClick, in a not too formal way so we could reference the version when need be.

So I propose to merge v1911 in a few days, and then create a tag. But I'm open to suggestions.

Tom


tbarbette:
Due to a false manipulation I had to reopen it at #216

PR #216, opened 2019-11-24
FastClick v19.11

tbarbette:
Hi all,

After solving a huge source of cache miss in DPDK TX, I thought it would be nice to version a bit FastClick, in a not too formal way so we could reference the version when need be.

So I propose to merge v1911 in a few days, and then create a tag. But I'm open to suggestions.

Tom

### DIFF
--- a/elements/standard/averagecounter.cc
+++ b/elements/standard/averagecounter.cc
@@ -24,23 +24,16 @@
 #include <click/error.hh>
 CLICK_DECLS
 
-AverageCounter::AverageCounter()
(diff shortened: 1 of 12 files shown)


gkatsikas:
Tagging FastClick sounds a good idea to me. I would say we should introduce one tag per year covering major improvements and/or bug fixes. Each tag should have a clear description of its features so users get a handy documentation of what is new.

tbarbette:
Probably more than years so we do not habe too much work  :p

PR #220, opened 2020-01-21
Towards DPDK 20.02

gkatsikas:
Compliance with DPDK 20.02-rc1


tbarbette:
Great! I actually needed it exactly right now :p

gkatsikas:
Good timing :D
Let's keep the PR alive until the final 20.02.

gkatsikas:
dpdk.md is getting very complex, we should do something about it

tbarbette:
I already added the compatibility with the new meson build system of DPDK (and linking with libdpdk as OSes do). The Make system will be removed at some point and at that point we'll stop changing that file.

PR #221, opened 2020-01-21
New ARPAdvertiser element

gkatsikas:
Input element that generates periodic ARP
responses to designated nodes.
It uses an internal ARP table which can be
populated with entries through a write handler.
Write handlers are also available for deletion
and cleaning of the ARP table.
This element is also batch compatible.

### DIFF
--- a/elements/ethernet/arpadvertiser.cc
+++ b/elements/ethernet/arpadvertiser.cc
@@ -1,0 +1,270 @@
+/*
+ * arpadvertiser.{cc,hh} -- ARP advertiser element
+ * Georgios Katsikas
+ *
+ * Copyright (c) 2020 UBITECH and KTH Royal Institute of Technology
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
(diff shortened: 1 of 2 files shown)


tbarbette:
This generates gratuitous ARP messages (still replies technically if I'm correct), right? Maybe it should be documented around that.

Why not using the ARPTable internally like the other ARP elements?

And then isn't it possible to make the ARPQuerier element trigger such message?

gkatsikas:
Basically, this element is an extension of the ARPFaker element.
The extension is about the ability to advertise more than one IP-MAC mappings to any host.
The reason that the ARPTable cannot be used is that the ARPAdvertiser element requires a tuple of the form: DstIP DstMAC AdvIP AdvMAC, while the ARP table can only store a single IP-MAC pair.

With additional effort, this element could be incorporated into the ARPResponder element, but I would not over-complicate that basic element.

tbarbette:
Okay for me then ;)

PR #223, opened 2020-01-22
Extensions & bug fixes for Check*Header elements

gkatsikas:
* atomic uint64_t counters (instead of uint32_t)
* More and better handlers for all elements
* Fixed uninitialized drop counters for CheckARPHeader CheckICMPHeader
* Batching support for CheckIPHeader, CheckIP6Header, CheckGridHeader
* Style changes


gkatsikas:
The work is finished here, feel free to provide review before eventually merging

tbarbette:
I saw multiple occurrences of output_push(), which will not work with the SimpleElement template, can you confirm it's not the case? Did you also check you only call push in non-batch mode and push_batch in batch mode?

tbarbette:
Slight correction: there is no problem using push or push_batch with the SimpleElement as long as the function returns 0 if the packet has been handled differently. However, please double check the in_batch_mode and batch mode.

gkatsikas:
All of these elements use simple_action and simple_action_batch.
I checked with batching disabled and printed 'in_batch_mode'. All of these elements report 0, which is BATCH_MODE_NO. In this case, only non-batching methods are called for processing (i.e., simple_action).
When I re-enable batching, all of them report BATCH_MODE_YES and simple_action_batch is invoked

elements/ethernet/checkarpheader.cc
```diff
@@ -60,24 +64,28 @@ CheckARPHeader::configure(Vector<String> &conf, ErrorHandler *errh)
...
         click_chatter("ARP header check failed: %s", reason_texts[reason]);
+    }
     _drops++;
 
-    if (_reason_drops)
+    if (_reason_drops) {
         _reason_drops[reason]++;
+    }
 
     checked_output_push(1, p);
```
tbarbette: This one is problematic. Under batch mode it will send a single packet as non-batch and lead to a crash.

tbarbette: But then it's not your fault, it was already there :p Can you just fix it?

gkatsikas: Yes, I will fix it

gkatsikas:
No more commits on my side, unless there is something to fix

tbarbette:
Great, thanks !

PR #224, opened 2020-01-23
Batch compatibility for 6 Ethernet elements

gkatsikas:

### DIFF
--- a/elements/ethernet/arpfaker.cc
+++ b/elements/ethernet/arpfaker.cc
@@ -2,7 +2,11 @@
  * arpfaker.{cc,hh} -- ARP response faker element
  * Robert Morris
  *
+ * Computational batching support
+ * by Georgios Katsikas
+ *
  * Copyright (c) 1999-2000 Massachusetts Institute of Technology
+ * Copyright (c) 2020 UBITECH and KTH Royal Institute of Technology
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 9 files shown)


tbarbette:
Seems good, thanks !

elements/ethernet/arpfaker.cc
```diff
@@ -57,9 +63,18 @@ ARPFaker::initialize(ErrorHandler *)
...
 {
-    if (Packet *p = ARPResponder::make_response(_eth1.data(), _ip1.data(),
-						_eth2.data(), _ip2.data()))
-	output(0).push(p);
+    Packet *p = ARPResponder::make_response(
+        _eth1.data(), _ip1.data(), _eth2.data(), _ip2.data()
+    );
+
+    if (p) {
+    #if HAVE_BATCH
```
tbarbette: This is not sufficient, as a path may be not batch-enabled, there is the in_batch_mode() function to check that

elements/ethernet/etherpausesource.cc
```diff
@@ -97,12 +104,18 @@ EtherPauseSource::cleanup(CleanupStage)
...
 EtherPauseSource::run_timer(Timer *)
 {
-    if (Packet *p = _packet->clone()) {
-	++_count;
-	output(0).push(p);
+    Packet *p = _packet->clone();
+    if (p) {
+        ++_count;
+    #if HAVE_BATCH
+        output_push_batch(0, PacketBatch::make_from_packet(p));
```
tbarbette: Same here

gkatsikas: Sure I can replace:
#if HAVE_BATCH
with
if (in_batch_mode == BATCH_MODE_YES) {

but I believe it is not necessary in this case.
Both elements are source elements (i.e., they have on input ports) and both of them have:
#if HAVE_BATCH
    in_batch_mode = BATCH_MODE_YES;
#endif

in their constructor, so the path begins from these elements and in_batch_mode is always BATCH_MODE_YES.

If you still believe I am wrong, I can push the changes :)

tbarbette: You are correct. I forgot they were source and therefore always in batch mode.

PR #225, opened 2020-01-28
Batching support and active param for IP6Print

gkatsikas:
Also, style changes for IP6Print and IPPrint
and documentation for IP6Print


tbarbette:
Seems good to me, thanks !

PR #226, opened 2020-01-28
Flow subsystem

tbarbette:
Merge the flow subsystem, inspired from MiddleClick

These commits bring the idea of separating the Flow classification from the stateful functions. Eg, packets go through FlowIPManager that will ensure some space is allocated for all flows passing by. Then one can put a FlowIPNat, and a FlowHyperscan to do a NAT and DPI. The advantage is that we have multiple implementations of the FlowIPManager that can be tailored to the use-case, while stateful elements get much simpler.

Also, the flow space assignation is done as per MiddleClick algo. All stateful elements are traversed to build a single flow table, with enough space for each of the elements.

It proved useful, both for RSS++, where a special FlowIPManager can take care of the migration and for Cheetah where by using the CheetahStateful element, one can directly use all the provided Flow\* elements, such as the IPLoadBalancer, at the core of the paper.

The flows are managed through a FlowControlBlock, a pointer to it is allocated on the thread stack. Batches going through the graph should be batches of the same session, hence the same control block.

What is not taken from MiddleClick (yet?) is the ageing system, the classification itself (the tree etc), the context links, or the whole idea of context and in-the-middle stream modification.

Note that some *research* Flow\*Manager\* have been done by the CSD students, such as a beginning of a hardware version of FlowIPManager.

There exists a few FlowIPManager as of now, all for TCP only:
 * FlowIPManager a DPDK-cuckoo based TCP flow manager
 * FlowIPManagerMP, same just with the MP flags set on, can, therefore, be accessed by multiple threads
 * FlowIPManagerSpinLock, the first one but protected by a SpinLock, just for reasearch purpose, prefer the second one.
 * FlowIPManagerIMP, the first one but duplicated per-core, no ageing yet. Faster than the second, but needs to carefully think how to handle the "returning" side.
 * FlowIPManagerHMP, based on the HashtableMP, no ageing yet
 * FlowIPManagerHW, in the research folder, does not work.


bcronje:
Hi Tom. This seems like a very interesting and useful commit. Can you confirm if my high-level understanding is correct:

FlowIPManager will be responsible to keep state of flows, e.g. keep a hash table or similar of all flows and responsible of adding new and removing old flows.
Stateful elements will somehow register with FlowIPManager to indicate it needs a portion of allocated state space.
FlowIPManager will allocate enough state space per flow packet so each registered stateful element will have their own portion of state available.

Is that more or less correct from a very high-level?
Is FlowControlBlock accessible by stateful elements on a per packet basis?
Does FlowControlBlock expose a function for a stateful element to get a pointer to the element's specific state?

tbarbette:
Yes you have the overall view 100% correct.

At initialization time, the FlowIPManager visits all downstream elements, asking for their space requirement (a virtual function call, but I provide helpers to define a structure that should be per-flow, so that is automatic most of the time), and then after computing the FCB layout, it informs each element of the offset where to find the space it asked for in the FCB (VirtualFlowSpaceElement::_fcb_offset). A template is available for convenience that hides all of that:
```
class MyElement : public FlowSpaceElement<**uint32_t**> {
    void push_batch(int port, **uint32_t*** perflow, PacketBatch* batch) {

    }
}
```

If the FlowIPManager receives a batch of 32 packets, it will break it in sub-batches of packets of the same flow. So if the batch starts with 3 packets of session A, it will send further down a batch of 3 packets. For everything done in the push path, the "FlowControlBlock *fcb_stack" pointer will be set to the FlowControlBlock of that session A. One could have passed "fcb_stack" with the batch, modifying all push_batch functions of FastClick, which was not done because a) it would mean changing all elements b) the FCB is not expected to change.
On the other hand, the method does not work with queue and thread switching. This is actually fine as most perf-flow stuffs should be done in one thread, eventually passing off packets to a queue that does not really need to know from which flow each packet comes anymore.

bcronje:
Great, thank you for the explanation. I think this is pretty neat!

elements/research/flowipmanagerhmp.cc
```diff
@@ -0,0 +1,124 @@
+/*
+ * FlowIPManagerHMP.{cc,hh}
```
gkatsikas: Description and copyrights

elements/research/flowipmanagerhmp.hh
```diff
@@ -0,0 +1,68 @@
+#ifndef CLICK_FlowIPManagerHMPMP_HH
```
gkatsikas: Use upper case for consistency

elements/research/flowipmanagerspinlock.cc
```diff
@@ -0,0 +1,209 @@
...
+    }
+};
+
+void FlowIPManagerSpinlock::add_handlers() {
+
+}
+
+CLICK_ENDDECLS
+
+EXPORT_ELEMENT(FlowIPManagerSpinlock)
```
gkatsikas: Element requires userlevel and dpdk?

tbarbette:
I'll merge this, and we'll make adaptations as time goes. Eg we need further work on ageing (and show how outdated the heap in the Rewriter elements are)

PR #227, opened 2020-01-29
Display message about batching in configure

gkatsikas:

### DIFF
--- a/configure
+++ b/configure
@@ -807,6 +807,7 @@ enable_batch
 enable_verbose_batch
 enable_auto_batch
 enable_netmap_pool
+enable_json
 enable_select
 enable_poll
 enable_kqueue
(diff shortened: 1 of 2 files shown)


tbarbette:
It will be overwritten if it's not done in configure.in first. I would suggest doing it in the .in, then do an autoreconf and send both changes.

gkatsikas:
> It will be overwritten if it's not done in configure.in first. I would suggest doing it in the .in, then do an autoreconf and send both changes.

Oh, you are right, I forgot that.
I re-post asap

gkatsikas:
Apparently, configure.in was missing more features than the batching flag.

tbarbette:
Yes I'm not surprised of that... :p Thanks !

PR #230, opened 2020-02-03
Fixed flow elements' dependencies

gkatsikas:
These dependencies manifest themselves when DPDK is not present


tbarbette:
Thanks !

PR #244, opened 2020-04-06
Reworked initialize_device()

gkatsikas:
An issue before this patch was that rte_eth_dev_configure
was invoked before the offloading capabilites were set.
Now, all the configuration is applied and rte_eth_dev_start
is moved at the end of the initialize_device method.


tbarbette:
Thanks!

PR #248, opened 2020-04-23
Use march=native by default, and check for AVX2

tbarbette:
Add the Search element that searchs for a pattern in the packet, using AVX2 if available. Hence all the importance of AVX2.

At the same time I think we should have -march=native by default (it was probably already added by DPDK anyway). One can disable it with --enable-portable-binary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ script:
       FRAMEWORK_FLAGS="--enable-user-multithread";
     fi
 
-  - ./configure $CONFIG $FLAGS $FRAMEWORK_FLAGS && make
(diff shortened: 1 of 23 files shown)


aliireza:
If the problem is only AVX2 instructions, you can still keep the -march=native, but disable AVX2:
-march=native -fno-avx2 

Check here for more flags:
https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/i386-and-x86-64-Options.html

tbarbette:
Thanks for your comment. My intent for AVX was to enable auto-detection by default to use in specific algorithms, or disable the explicit usage of AVX2. But what the compiler does is another mater. The user can still pass CXXFLAGS=-fno-avx2 if she wants.

PR #251, opened 2020-04-28
New script handlers

tbarbette:
Somehow I forgot this branch :p

### DIFF
--- a/elements/standard/script.hh
+++ b/elements/standard/script.hh
@@ -321,6 +321,18 @@ numbers; for example, 'C<sub 10 5 2>' returns
 Finds the minimum or maximum of a space-separated list of
 numbers; for example, 'C<max 5 10 2>' returns "C<10>".
 
+=h avg "read with parameters"
+
+Finds the average of a space-separated list of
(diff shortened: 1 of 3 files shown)


elements/standard/script.hh
```diff
@@ -321,6 +321,18 @@ numbers; for example, 'C<sub 10 5 2>' returns
...
+Finds the average of a space-separated list of
+numbers; for example, 'C<avg 5 10 2>' returns "C<6>".
+
+=h count, popcount "read with parameters"
+
+Finds the number of items, or the numver of items different than zero
+of a space-separated list of
+numbers; for example, 'C<count 5 0 2>' returns "C<3>" and
+'C<popcount 5 0 2>' returns "C<2>".
+
```
gkatsikas, line 335: Include documentation for gt_vec

gkatsikas: My bad, I did not see the other commits, gt_vec documentation is in place

tbarbette: Yeah no it's me, I did a second pass just after the pull request

PR #253, opened 2020-04-28
DDIOTune element added

aliireza:
DDIOTune element can enable/disable DDIO and change the number of IIO LLC ways. Besides, it can print some info about the cache hierarchy. Let me know if I need to change anything.

### DIFF
--- a/elements/research/ddiotune.cc
+++ b/elements/research/ddiotune.cc
@@ -29,353 +29,325 @@
 CLICK_DECLS
 
 /* Read from an MSR register */
-int32_t
-DDIOTune::rdmsr(int msr, uint64_t *val, int cpu)
-{
-	int fd;
-	int ret;
-	char msr_file[50];
-	sprintf(msr_file,"/dev/cpu/%d/msr",cpu);
+int DDIOTune::rdmsr(int msr, uint64_t *val, int cpu) {
+  int fd;
+  char msr_file[50];
+  sprintf(msr_file, "/dev/cpu/%d/msr", cpu);
 
-	fd = open(msr_file, O_RDONLY);
-	if (fd < 0)
-		return fd;
+  fd = open(msr_file, O_RDONLY);
+  if (fd < 0)
+    return fd;
 
-	ret = pread(fd, val, sizeof(uint64_t), msr);
+  int ret = pread(fd, val, sizeof(uint64_t), msr);
 
-	close(fd);
+  close(fd);
 
-	return ret;
+  return ret;
 }
 
-
 /* Write to an MSR register */
-int32_t
-DDIOTune::wrmsr(int msr, uint64_t *val, int cpu)
-{
-	int fd;
-	int ret;
-	char msr_file[50];
-	sprintf(msr_file,"/dev/cpu/%d/msr",cpu);
+int DDIOTune::wrmsr(int msr, uint64_t *val, int cpu) {
+  int fd;
+  char msr_file[50];
+  sprintf(msr_file, "/dev/cpu/%d/msr", cpu);
 
-	fd = open(msr_file, O_WRONLY);
-	if (fd < 0)
-		return fd;
+  fd = open(msr_file, O_WRONLY);
+  if (fd < 0)
+    return fd;
 
-	ret = pwrite(fd, val, sizeof(uint64_t), msr);
+  int ret = pwrite(fd, val, sizeof(uint64_t), msr);
 
-	close(fd);
+  close(fd);
 
-	return ret;
+  return ret;
 }
 
 /* Initialize PCI access data structure */
-void
-DDIOTune::init_pci_access(void)
-{
-	pacc = pci_alloc();           /* Get the pci_access structure */
-	pci_init(pacc);               /* Initialize the PCI library */
-	pci_scan_bus(pacc);           /* We want to get the list of devices */
+int DDIOTune::init_pci_access() {
+  pacc = pci_alloc(); /* Get the pci_access structure */
+  if (pacc == NULL)
+    return -1;
+  pci_init(pacc);     /* Initialize the PCI library */
+  pci_scan_bus(pacc); /* We want to get the list of devices */
+  return 0;
 }
 
 /* Find the responsible PCI root for the input NIC */
-struct pci_dev*
-DDIOTune::find_ddio_device(uint8_t nic_bus)
-{
-	struct pci_dev* dev;
-	for(dev = pacc->devices; dev; dev=dev->next) {
-		pci_fill_info(dev,PCI_FILL_IDENT | PCI_FILL_BASES | PCI_FILL_NUMA_NODE | PCI_FILL_PHYS_SLOT);
-		/* 
-		 * Find the proper PCIe root based on the nic device
-		 * For instance, if the NIC is located on 0000:17:00.0 (i.e., BDF)
-		 * 0x17 is the nic_bus (B)
-		 * 0x00 is the nic_device (D)
-		 * 0x0	is the nic_function (F)
-		 */
-		if(/*dev->func == 0 && dev->dev == 0  && */pci_read_byte(dev,PCI_SUBORDINATE_BUS) == nic_bus ) {
-		return dev;
-		}
-	}
-	printf("Could not find the proper PCIe root!\n");
-	return NULL;
+struct pci_dev *DDIOTune::find_ddio_device(uint8_t nic_bus) {
+  struct pci_dev *dev;
+  for (dev = pacc->devices; dev; dev = dev->next) {
+    pci_fill_info(dev, PCI_FILL_IDENT | PCI_FILL_BASES | PCI_FILL_NUMA_NODE |
+                           PCI_FILL_PHYS_SLOT);
+    /*
+     * Find the proper PCIe root based on the nic device
+     * For instance, if the NIC is located on 0000:17:00.0 (i.e., BDF)
+     * 0x17 is the nic_bus (B)
+     * 0x00 is the nic_device (D)
+     * 0x0	is the nic_function (F)
+     */
+    if (pci_read_byte(dev, PCI_SUBORDINATE_BUS) == nic_bus) {
+      return dev;
+    }
+  }
+  return NULL;
 }
 
-
 /* Print cache information */
-void
-DDIOTune::print_cache_info()
-{
-	click_chatter("Cache Hierarchy Info");
-	click_chatter("====================");
-	for(int i=0;i<CACHE_LEVELS;i++){
-		click_chatter("Level %d:",i);
-		click_chatter("#Ways: %d",ways[i]);
-		//click_chatter("#Partitions: %"PRIu32,partitions[i]);
-		//click_chatter("Cache Line Size: %d",linesize[i]);
-		click_chatter("#Sets: %d",sets[i]);
-		click_chatter("Size: %d B",size[i]);
-		click_chatter("====================");
-	}
-
+void DDIOTune::print_cache_info() {
+  click_chatter("Cache Hierarchy Info");
+  click_chatter("====================");
+  for (int i = 0; i < CACHE_LEVELS; i++) {
+    click_chatter("Level %d:", i);
+    click_chatter("#Ways: %d", ways[i]);
+    click_chatter("#Partitions: %" PRIu32, partitions[i]);
+    click_chatter("Cache Line Size: %d", linesize[i]);
+    click_chatter("#Sets: %d", sets[i]);
+    click_chatter("Size: %d B", size[i]);
+    click_chatter("====================");
+  }
 }
 
 /* Check DDIO status (i.e., enabled or disabled) for the input NIC */
-int
-DDIOTune::ddio_status(uint8_t nic_bus)
-{
-	uint32_t val;
-	if(!pacc)
-		init_pci_access();
-
-	struct pci_dev* dev=find_ddio_device(nic_bus);
-	if(!dev){
-		printf("No device found!\n");
-		printf("No device found!\n");
-		exit(1);
-	}
-	/*
-	 * perfctrlsts_0
-	 * bit 3: NoSnoopOpWrEn -> Should be 1b
-	 * bit 7: Use_Allocating_Flow_Wr -> Should be 0b
-	 * Check p. 68 of Intel® Xeon® Processor Scalable Family
-	 * Datasheet, Volume Two: Registers
-	 * May 2019
-	 * link: https://www.intel.com/content/www/us/en/processors/xeon/scalable/xeon-scalable-datasheet-vol-2.html
-	 * Haswell link: https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/xeon-e5-v2-datasheet-vol-2.pdf
-	 */
-	val=pci_read_long(dev,PERFCTRLSTS_0);
-	printf("perfctrlsts_0 val: 0x%" PRIx32 "\n",val);
-	printf("NoSnoopOpWrEn val: 0x%" PRIx32 "\n",val&nosnoopopwren_MASK);
-	printf("Use_Allocating_Flow_Wr val: 0x%" PRIx32 "\n",val&use_allocating_flow_wr_MASK);
-	if(val&use_allocating_flow_wr_MASK)
-		return 1;
-	else
-		return 0;
+int DDIOTune::ddio_status() {
+  /*
+   * perfctrlsts_0
+   * bit 3: NoSnoopOpWrEn -> Should be 1b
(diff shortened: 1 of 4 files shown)


elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+/* Read from an MSR register */
+int32_t
+DDIOTune::rdmsr(int msr, uint64_t *val, int cpu)
+{
+	int fd;
+	int ret;
+	char msr_file[50];
+	sprintf(msr_file,"/dev/cpu/%d/msr",cpu);
+
+	fd = open(msr_file, O_RDONLY);
```
gkatsikas, line 40: Instead of issuing 2 syscalls (open and pread), you could simply use the native Click method: file_string()
This method returns a String, but you could easily convert the result to int.

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+CLICK_DECLS
+
+/* Read from an MSR register */
+int32_t
+DDIOTune::rdmsr(int msr, uint64_t *val, int cpu)
+{
+	int fd;
+	int ret;
+	char msr_file[50];
+	sprintf(msr_file,"/dev/cpu/%d/msr",cpu);
```
gkatsikas, line 38: Make sure you leave a space after each argument (e.g., msr_file, "..", cpu)

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	close(fd);
+
+	return ret;
+}
+
+/* Initialize PCI access data structure */
+void
+DDIOTune::init_pci_access(void)
+{
+	pacc = pci_alloc();           /* Get the pci_access structure */
```
gkatsikas, line 76: What if pacc == NULL?
Shouldn't you return with an error (turn void to int)?

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	pci_init(pacc);               /* Initialize the PCI library */
```
gkatsikas, line 77: Is this method void or does it return a status? If a status is returned you need to check if it succeeded and only then proceed to the next call.

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	pci_scan_bus(pacc);           /* We want to get the list of devices */
```
gkatsikas, line 78: Same here

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	for(dev = pacc->devices; dev; dev=dev->next) {
+		pci_fill_info(dev,PCI_FILL_IDENT | PCI_FILL_BASES | PCI_FILL_NUMA_NODE | PCI_FILL_PHYS_SLOT);
+		/* 
+		 * Find the proper PCIe root based on the nic device
+		 * For instance, if the NIC is located on 0000:17:00.0 (i.e., BDF)
+		 * 0x17 is the nic_bus (B)
+		 * 0x00 is the nic_device (D)
+		 * 0x0	is the nic_function (F)
+		 */
+		if(/*dev->func == 0 && dev->dev == 0  && */pci_read_byte(dev,PCI_SUBORDINATE_BUS) == nic_bus ) {
```
gkatsikas, line 95: Better remove this comment

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+		return dev;
```
gkatsikas, line 96: This line needs proper alignment

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	pacc = pci_alloc();           /* Get the pci_access structure */
+	pci_init(pacc);               /* Initialize the PCI library */
+	pci_scan_bus(pacc);           /* We want to get the list of devices */
+}
+
+/* Find the responsible PCI root for the input NIC */
+struct pci_dev*
+DDIOTune::find_ddio_device(uint8_t nic_bus)
+{
+	struct pci_dev* dev;
```
gkatsikas, line 85: You can declare this variable at the beginning of the for loop

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+}
+
+
+/* Print cache information */
+void
+DDIOTune::print_cache_info()
+{
+	click_chatter("Cache Hierarchy Info");
+	click_chatter("====================");
+	for(int i=0;i<CACHE_LEVELS;i++){
```
gkatsikas, line 110: Leave space before and after a semicolon

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+DDIOTune::ddio_status(uint8_t nic_bus)
+{
+	uint32_t val;
+	if(!pacc)
+		init_pci_access();
+
+	struct pci_dev* dev=find_ddio_device(nic_bus);
+	if(!dev){
+		printf("No device found!\n");
+		printf("No device found!\n");
```
gkatsikas, line 133: Why is this printed twice?

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+		exit(1);
```
gkatsikas, line 134: Exiting like this here is not the most graceful way to kill the router.
Better return a failure and call assert at the caller

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	 * perfctrlsts_0
+	 * bit 3: NoSnoopOpWrEn -> Should be 1b
+	 * bit 7: Use_Allocating_Flow_Wr -> Should be 0b
+	 * Check p. 68 of Intel® Xeon® Processor Scalable Family
+	 * Datasheet, Volume Two: Registers
+	 * May 2019
+	 * link: https://www.intel.com/content/www/us/en/processors/xeon/scalable/xeon-scalable-datasheet-vol-2.html
+	 * Haswell link: https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/xeon-e5-v2-datasheet-vol-2.pdf
+	 */
+	val=pci_read_long(dev,PERFCTRLSTS_0);
```
gkatsikas, line 146: spaces needed before and after the =

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+/* Enable DDIO for the input NIC */
+void
+DDIOTune::ddio_enable(uint8_t nic_bus)
+{
+	uint32_t val;
+	if(!pacc)
+		init_pci_access();
+
+	if(!ddio_status(nic_bus))
+	{
```
gkatsikas, line 165: This bracket should go at the same line with the if condition

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+		struct pci_dev* dev=find_ddio_device(nic_bus);
+		if(!dev){
+			printf("No device found!\n");
+			exit(1);
+		}
+		val=pci_read_long(dev,PERFCTRLSTS_0);
+		pci_write_long(dev,PERFCTRLSTS_0,val|use_allocating_flow_wr_MASK);
+		printf("DDIO is enabled!\n");
+	} else
+	{
```
gkatsikas, line 175: Same here, goes with else

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	uint32_t val;
+	if(!pacc)
+		init_pci_access();
+
+	if(ddio_status(nic_bus))
+	{
+		struct pci_dev* dev=find_ddio_device(nic_bus);
+		if(!dev){
+			printf("No device found!\n");
+			exit(1);
```
gkatsikas, line 193: Same issue with exit, make this method return int and assert at the caller

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	{
+		printf("DDIO was already disabled\n");
+	}
+}
+
+/* Print PCI root information */
+void
+DDIOTune::print_dev_info(struct pci_dev *dev)
+{
+	if(!dev){
```
gkatsikas, line 208: space before {

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	_numa_node=-1;
+	_msr_cpu=-1;
+	IIO_LLC_WAYS_VALUE_NEW=0;
+	ways = new uint32_t[CACHE_LEVELS];
+	partitions = new uint32_t[CACHE_LEVELS];
+	linesize = new uint32_t[CACHE_LEVELS];
+	sets = new uint32_t[CACHE_LEVELS];
+	size = new uint32_t[CACHE_LEVELS];
+
+	init_pci_access();
```
gkatsikas, line 236: this should go to the initialize() method of the element

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	/* Get IIO_LLC_WAYS info */
+	max_w=ways[CACHE_LEVELS-1];
+    IIO_LLC_WAYS_VALUE_MIN = 6<<(max_w - (max_w%4)); /* The first hex digit is always 6 (i.e., Skylake: 0x600 and Haswell: 0x60000). */
+	IIO_LLC_WAYS_VALUE_MAX = pow(2,max_w)-1;
+
+    //click_chatter("Min: 0x%x Max: 0x%x", IIO_LLC_WAYS_VALUE_MIN, IIO_LLC_WAYS_VALUE_MAX);
+
+	/* Read IIO LLC WAYS */
```
gkatsikas, line 329: This is block should also go to the initialize() method

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+		if (getuid()){
+			return errh->error("Need root access! Run again with sudo.");
+		}
+		else{
+			return errh->error("Make sure that msr module is loaded! Run: sudo modprobe msr");
+		}
+	}
+
+	
+	if(ddio_status(_pci_dev)) {
```
gkatsikas, line 343: Same here, initialize()

elements/research/ddiotune.hh
```diff
@@ -0,0 +1,127 @@
...
+
+    const char *class_name() const      { return "DDIOTune"; }
+    const char *port_count() const      { return PORTS_0_0; }
+
+    int configure_phase() const { return CONFIGURE_PHASE_FIRST; }
+
+    int configure(Vector<String>&, ErrorHandler*) override;
+
+    //TODO: One instance per pci dev
+    //static DDIOTune *instance;
```
gkatsikas, line 98: Check the FlowDispatcher element in the metron branch.
That element manages the HW Flow API for each NIC. I had the same issue. I wanted to manage all NIC with the same element. To do so, I used a an internal HashTable<portid_t, FlowDispatcher *> map.
You could use a static HashTable<portid_t, DDIOTune *>

elements/research/ddiotune.cc
tbarbette: That would lead to multiple conversion and go through unnecessary null-termination. And the file content would appear as a weird quad word as string. If anything then add a function like file_read_uint64. But it's so specific that I think it's fine.

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+
+    /* Print cache and PCI root information */
+	if(_print) {
+		print_cache_info();
+		print_dev_info(_dev);
+	}
+
+	/* Get IIO_LLC_WAYS info */
+	max_w=ways[CACHE_LEVELS-1];
+    IIO_LLC_WAYS_VALUE_MIN = 6<<(max_w - (max_w%4)); /* The first hex digit is always 6 (i.e., Skylake: 0x600 and Haswell: 0x60000). */
```
tbarbette, line 324: Space

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+		click_chatter("DDIO state is reverted back to disabled!");
+	}
+	pci_cleanup(pacc);		/* Close everything */
+}
+
+
+/* Configure: main function */
+int
+DDIOTune::configure(Vector<String> &conf, ErrorHandler *errh)
+{
```
tbarbette, line 266: Why this removal? But if you uncomment don't forget the instance(0) in the constructor.

elements/research/ddiotune.cc
```diff
@@ -0,0 +1,385 @@
...
+		return errh->error("N_WAYS cannot be %d. It should be greater than 2 and smaller than %d (i.e. number of LLC ways).", _w, max_w);
+
+	if(_w == 0) {
+		//ddio_disable(_dev_addr->bus);
+		ddio_disable(_pci_dev);
+		return 0;
+	} else {
+		//ddio_enable(_dev_addr->bus);
+		ddio_enable(_pci_dev);
+    	//Convert _w to the proper hexadecimal value to write
```
tbarbette, line 365: Space

elements/research/ddiotune.hh
```diff
@@ -0,0 +1,127 @@
...
+
+DDIOTune(N_WAYS, DEV [,PRINT])
+
+=s test
+
+Tune the DDIO capacity based on the input number of ways.
+
+=d 
+
+Configure IIO_LLC_WAYS register in Intel processors. In addition, this element can disable DDIO for a specific NIC.
```
tbarbette, line 24: It would be great to add a link to more information (not yet available I guess) and like 3 lines about what should be the parameters in which situations. And also 2 lines of context about DDIO and ways.

FYI, this element doc will automatically go to the wiki once I generate the doc (from time to time).

elements/research/ddiotune.cc
aliireza: I followed a typical way that they use these functions. However, I will add an assert to check the pointer.

elements/research/ddiotune.cc
aliireza: They do not return any status, see https://fossies.org/linux/pciutils/lib/pci.h

elements/research/ddiotune.cc
aliireza: If @tbarbette is okay with it, I will skip this :-p 
It will be called during the initialization phase, so 2 syscall is not a big issue.

elements/research/ddiotune.cc
aliireza: I will fix them. Forget to run reindent ;)

elements/research/ddiotune.cc
aliireza: Same as the other ones, no status.

elements/research/ddiotune.cc
aliireza: Will do it.

elements/research/ddiotune.cc
aliireza: Will do it.

elements/research/ddiotune.cc
aliireza: Will do it.

elements/research/ddiotune.cc
aliireza: Typo :-p

elements/research/ddiotune.cc
aliireza: Reindent will fix this

elements/research/ddiotune.cc
aliireza: Reindent will fix this

elements/research/ddiotune.cc
bcronje: Shouldn't `click_chatter` be preferred over `sprintf` ?

elements/research/ddiotune.cc
aliireza: I will move it to the configure method. Would it be okay?

elements/research/ddiotune.cc
aliireza: I fixed it based on Georgios's comment.

elements/research/ddiotune.hh
aliireza: I will add more info ;)

elements/research/ddiotune.hh
aliireza: Thanks for the pointer. I fixed it.

aliireza:
I think I fixed most of the changes that you asked. Can you check?

tbarbette:
Good for me ! I'll merge when Travis finish :)

tbarbette:
Thanks!

aliireza:
> Thanks! Please add your paper in the "community" page of the wiki when it's time :)

Thanks. I will ;)

PR #255, opened 2020-05-01
Merge Fastclick in Metron

tbarbette:
Might be done a little bit quickly. To be checked again. The deactivation of enable-flow-api might not work (I wanted that for RSS++ too).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ cache:
     - dpdk-18.08
     - dpdk-19.08
     - dpdk-19.11
+    - dpdk-20.02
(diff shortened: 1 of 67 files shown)


gkatsikas:
I believe we can merge at this point.

tbarbette:
Actually RSS++ merge was a mistake on my end, that part is not in so the flag does not make sense yet...

tbarbette:
I cleaned the history mentioning your change in the merge commit and merged the last fix, I had skipped one of your comments it seems.

gkatsikas:
Looks good to me!

PR #256, opened 2020-05-06
Optionally print MAC addresses with IPPrint

gkatsikas:


gkatsikas:
@tbarbette can you initiate travis?
It looks like I cannot do it myself.

tbarbette:
Something is weird. I can't either.

tbarbette:
Did you test "make check" ? The IPPrint change may cause tests to fail. Better to wait for travis...

tbarbette:
Though as the default is to not change anything it should not be problematic...

gkatsikas:
> Did you test "make check" ? The IPPrint change may cause tests to fail. Better to wait for travis... Else I can set up a gitlab-ci at KTH.

Yes, I tried make check, everything looks good with respect to IPPrint tests

gkatsikas:
If travis revoked your account, then it is a good idea to setup CI/CD at KTH

PR #257, opened 2020-05-22
FromDPDKDevice: Cleanup

tbarbette:
Remove load tracking, that is done with --enable-cpu-load
Put everything interrupt related under a #ifdef, as it is yet not tested

### DIFF
--- a/elements/userlevel/queuedevice.cc
+++ b/elements/userlevel/queuedevice.cc
@@ -491,22 +491,6 @@ unsigned long long QueueDevice::n_count() {
     return total;
 }
 
-unsigned long long QueueDevice::n_useful() {
-    unsigned long long total = 0;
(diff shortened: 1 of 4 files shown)


elements/userlevel/queuedevice.cc
```diff
@@ -491,22 +491,6 @@ unsigned long long QueueDevice::n_count() {
     return total;
 }
 
-unsigned long long QueueDevice::n_useful() {
```
gkatsikas, line 494: Aren't the methods n_useful() and n_useless() defined in queuedevice.hh?

gkatsikas: Apart from this, looks good to me

tbarbette:
Thanks for the review, indeed I forgot n_useful etc.

PR #258, opened 2020-05-23
FlowDispatcher: Rename to FlowParser

tbarbette:
Not finished, need to extract what is not "parser". FlowManager is taken. So FlowRuleManager? And maybe FlowRuleParser instead of just FlowParser.


configure
```diff
@@ -11069,15 +11069,15 @@ else
   ac_have_flow_disp_h=no
 fi
 
-as_ac_File=`$as_echo "ac_cv_file_$srcdir/include/click/flowdispatcherparser.hh" | $as_tr_sh`
-{ $as_echo "$as_me:${as_lineno-$LINENO}: checking for $srcdir/include/click/flowdispatcherparser.hh" >&5
-$as_echo_n "checking for $srcdir/include/click/flowdispatcherparser.hh... " >&6; }
+as_ac_File=`$as_echo "ac_cv_file_$srcdir/include/click/flowparserparser.hh" | $as_tr_sh`
```
gkatsikas, line 11072: This is funny :p You could rename it to flowparsersquare :p
This is why FlowRuleManager or NICRuleManager is more relevant. For now , leave it as it is and we will merge properly into master

include/click/flowparserglue.hh
```diff
@@ -20,6 +20,10 @@
 #ifndef CLICK_FLOWDISPATCHER_GLUE_H
```
gkatsikas, line 20: These names must be updated accordingly, but let's do it later

include/click/flowparserglue.hh
```diff
@@ -124,6 +128,77 @@ int cmdline_parse(
 
 #ifdef __cplusplus
 }
+
+#define FLOWDISP_ERROR   ((int)-1)
```
gkatsikas, line 132: Same here

configure
tbarbette: That's a mistake actually :p flowparserparser.hh does not exist ^^

gkatsikas:
I suggest to properly address the naming of this component now that there is no time pressure

tbarbette:
Yep! I'd say FlowParser is okay but we need to extract the managing part to FlowRuleManager. We can keep the system name as FlowParser because all in all, all that is for parsing the DPDK format. But that's your code, I'll follow ;)

gkatsikas:
More tomorrow :p

gkatsikas:
I am done with the refactoring, I believe we can merge if Travis agrees

gkatsikas:
> I am done with the refactoring, I believe we can merge if Travis agrees

Actually, there is more to do.
We could decouple the FlowCache into a separate file and rename it to FlowRuleCache.
The remaining part could be renamed to FlowRuleManager

gkatsikas:
Now, the floor is yours :)

tbarbette:
Great! Thanks!

PR #261, opened 2020-06-01
Reworked rule generators

gkatsikas:
* Object-oriented design
* New OpenFlow rule generator element
* Bug fixes related to generated port numbers
* New example configurations

### DIFF
--- a/elements/ip/generateipfilterrules.cc
+++ b/elements/ip/generateipfilterrules.cc
@@ -1,6 +1,6 @@
 // -*- c-basic-offset: 4; related-file-name: "generateipfilter.hh" -*-
 /*
- * generateipfilter.{cc,hh} -- element generates IPFilter patterns out of input traffic
+ * generateipfilterrules.{cc,hh} -- element generates IPFilter rule patterns out of input traffic
  * Tom Barbette, (extended by) Georgios Katsikas
  *
  * Copyright (c) 2017 Tom Barbette, University of Liège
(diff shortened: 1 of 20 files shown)


tbarbette:
Nice! Those needed rework. I did a quick lookup and it's fine by me.

gkatsikas:
I uncovered another bug apart. Once travis approves, I will merge

elements/ip/generateipfilterrules.cc
```diff
@@ -154,30 +154,30 @@ GenerateIPFilter::configure(Vector<String> &conf, ErrorHandler *errh)
...
+GenerateIPFilterRules::simple_action(Packet *p)
 {
     // Create a flow signature for this packet
     IPFlowID flowid(p);
     IPFlow new_flow = IPFlow();
     new_flow.initialize(flowid & _mask);
 
     // Check if we already have such a flow
-    IPFlow *found = _map.find(flowid).get();
+    IPFlow *found = _map.find(flowid & _mask).get();
```
gkatsikas, line 180: Don'y you think that this was a bug?
Line 177 initializes a flow with a flow ID which derives from flowid & mask, but then line 180 searches for an existing flow with flowid. Apparently this is not the same flow

tbarbette: What is sure is that new_flow should be in the if branch. And then yes we need & _mask in both side. Probably flowid = flowid & _mask as we won't use it unmasked.

PR #263, opened 2020-06-01
Batching support for RR sched elements

gkatsikas:
Along with style changes


gkatsikas:
As far as I tested yes.

gkatsikas:
> It works the same with and without batching right? It will pull in a per-packet RR ways N times for a batch of N packets, correct?

Actually before this patch we were using this element to generate traces for our last project, so batching support was not there and the element was using a regular pull(). I tested with a similar trace and it works

tbarbette:
Okay then! Thanks!

PR #268, opened 2020-09-02
FromDPDKDevice: avoid global enum definition

tbarbette:
I saw this way of doing things in multiple elements. Seems better to avoid names that could clash between inherited instances etc.


gkatsikas (approved):
Looks good to me!

PR #271, opened 2020-09-25
Improve handling of global DPDK in configure

skyfmmf:
With a global DPDK installation, the configure script showed an error
after the detection instead of "yes" and the check for read_clock
support was never run against the global DPDK.


tbarbette:
Thanks!

PR #273, opened 2020-10-07
Fix Meson/system build and warning

tbarbette:
Also add travis support. For now travis produce a problem with clang.
However it's not reproducible locally. I'll verify with the next build
if it still appears. It is in the exclude list for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,10 @@ cache:
   apt: true
   directories:
     - dpdk-16.04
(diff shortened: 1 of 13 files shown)


gkatsikas:
@tbarbette I will review it after October 23 ;)
This is still way before the DPDK 20.11 release :)

tbarbette:
I need this now for meson based install

PR #274, opened 2020-10-13
Checksum offloading for TX (UDP) and RX (IP, TCP, UDP) with DPDK

skyfmmf:
This allows to activate UDP checksum offloading for ToDPDKDevice, and allows to activate IP, TCP and UDP checksum offloading for FromDPDKDevice.


tbarbette:
Great! Thanks!

PR #275, opened 2020-10-15
Minor fixes and improvements

MassimoGirondi:
- Fix handlers for counters in FromDPDKDevice
- Fix count handler in FlowIPManagerIMP, now it returns the total number of flows
- Add `verbose` to FlowRuleManager to allow silent operations
- Solve division by zero error in CyccleCountAccum when packet count is 0

### DIFF
--- a/elements/userlevel/fromdpdkdevice.cc
+++ b/elements/userlevel/fromdpdkdevice.cc
@@ -405,7 +405,7 @@ void FromDPDKDevice::selected(int fd, int mask) {
 
 enum {
     h_vendor, h_driver, h_carrier, h_duplex, h_autoneg, h_speed, h_type,
-    h_ipackets, h_ibytes, h_imissed, h_ierrors, h_nombufs,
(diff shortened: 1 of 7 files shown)


elements/userlevel/fromdpdkdevice.cc
```diff
@@ -405,7 +405,7 @@ void FromDPDKDevice::selected(int fd, int mask) {
 
 enum {
     h_vendor, h_driver, h_carrier, h_duplex, h_autoneg, h_speed, h_type,
-    h_ipackets, h_ibytes, h_imissed, h_ierrors, h_nombufs,
+    h_ipackets, h_ibytes, h_imissed, h_ierrors, h_nombufs, h_out_of_buffer,
```
tbarbette, line 408: This is already possible with xstats I think, where you can give a name. "fd.xstats out_of_buffer", no? Because we could do this for all xstats.

MassimoGirondi: Ops, I missed that. I added an indication of that to the documentation.

PR #276, opened 2020-10-15
Atomics variables

MassimoGirondi:
- Add support for C++11 atomic types, which could be enabled with `--enable-atomic-builtins`.
- TSC counter for all DPDK supported platforms
- Minor additions to `.gitignore`

### DIFF
--- a/include/click/atomic.hh
+++ b/include/click/atomic.hh
@@ -28,6 +28,25 @@ CLICK_CXX_UNPROTECT
 # endif
 #endif
 
+
+#if HAVE_MULTITHREAD && HAVE_ATOMIC_BUILTINS
+//    #pragma message "atomic.hh: using builtins implementation"
+    # define CLICK_ATOMIC_BUILTINS 1
+    # define CLICK_ATOMIC_MEMORDER __ATOMIC_SEQ_CST
+    //	compare_swap methods are not atomic if using builtins. Disable them
+    # define CLICK_BUILTINS_DEPRECATED = delete
+    # define CLICK_BUILTINS_NONDEPRECATED
+    # define CLICK_ATOMIC_COMPARE_SWAP 0
+#else
+    # define CLICK_ATOMIC_BUILTINS 0
+//    #pragma message "atomic.hh: Using click own implementation"
+    // compare_swap are truly atomics if correctly implemented (e.g. x86).
+    // In this case, we maintain the original Click's deprecation flags
+    # define CLICK_BUILTINS_DEPRECATED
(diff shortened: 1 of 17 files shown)


include/click/atomic.hh
```diff
@@ -76,15 +95,16 @@ class atomic_uint32_t { public:
...
-    inline bool compare_and_swap(uint32_t expected, uint32_t desired) CLICK_DEPRECATED;
+    inline uint32_t compare_swap(uint32_t expected, uint32_t desired) CLICK_BUILTINS_DEPRECATED;
+    inline bool compare_and_swap(uint32_t expected, uint32_t desired) CLICK_BUILTINS_NONDEPRECATED;
 
     inline static uint32_t swap(volatile uint32_t &x, uint32_t desired);
     inline static void inc(volatile uint32_t &x);
     inline static bool dec_and_test(volatile uint32_t &x);
-    inline static uint32_t compare_swap(volatile uint32_t &x, uint32_t expected, uint32_t desired);
-    inline static bool compare_and_swap(volatile uint32_t &x, uint32_t expected, uint32_t desired) CLICK_DEPRECATED;
+    inline static uint32_t compare_swap(volatile uint32_t &x, uint32_t expected, uint32_t desired) CLICK_BUILTINS_DEPRECATED;
```
tbarbette, line 104: What is the one to use now then? :p

MassimoGirondi: The previous Click code used `uint32_t compare_swap`: https://github.com/tbarbette/fastclick/commit/4fdcf82c659fbff108906b6a97d48972e9b9291a.
However, it is fully atomic only on x86 implementation.

The rationale behind the flags and the various `#if CLICK_COMPARE_SWAP` around the code is to maintain compatibility:
- if `--enable-atomic-builtins` is given, `compare_swap` is marked as deleted and all the calls are replaced with the equivalent `compare_and_swap`
- by default, the old deprecated mark on `compare_and_swap` is maintained, while `compare_swap` is used in the calls.

I think that there isn't any usage in the code that requires `compare_swap` instead of `compare_and_swap` and all the occurrences could be replaced by `compare_and_swap`.

tbarbette:
Would be great to find a way to make it work with Travis :p

tbarbette:
Ok, travis was a random error apparently. Shall I merge now?

tbarbette:
@MassimoGirondi Did you forget to push the configure.in change? Can you PR it? Else the configure script is generated from configure.in, hence your changes will be lost when one touch configure.in or calls ./autogen.sh

MassimoGirondi:
Yes, my fault :/. Should be fine now.

Issue #254, opened 2020-04-29
Problems with --enable-flow-api


tbarbette:
I fast-pushed a commit to disable the hardware flow API by default, as there are issues caused by the dependency order still (tools/lib may compile before rte_parse is done, leading to "failed" warnings). Also with "-O1 -g" @aliireza could not compile due to some testpmd.h missing file.

Then I'm concerned the build will fail in two cases:
- When DPDK is "installed" and RTE_SDK is therefore a path to an installed DPDK, which should always be the case as per discussion with Thomas Montjalon
- When we don't use the DPDK make system but meson, which is the default in 20.08 I think, and only pkg-config should be used. I'm not even sure librte_parse sources can be found using that technique. 

Poke @gkatsikas

PR #278, opened 2020-10-26
Meson: fix in version check and librte_parse compilation

MassimoGirondi:
This is a fix that allow to compile fastclick with the flow_api with DPDK 20.08 and beyond.

When DPDK is found with `pkg-config`, this has the precedence on the RTE_SDK/RTE_TARGET system.

However, when `--enable-flow-api` is enabled, the `rte_parse` library needs to be compiled. Therefore, to identify the `testpmd` source files, the RTE_SDK variable is used (and must be set correctly).

Probably not the most elegant solution, but it works.

Also, the DPDK versions variable are now correctly populated when compiling with meson-installed DPDK.

The compilation of librte_parse library has been separated from the
`dpdk.mk` file so it can be included without re-defining all the dpdk
variables.

Related: https://github.com/tbarbette/fastclick/issues/254#issue-608947608


gkatsikas:
Good catch Massimo, I also noticed this issue recently, but didn't have the time to fix it.
+1 for me if you have already tested that it works

MassimoGirondi:
It seems to work, I tested with DPDK 20.02 and 20.08.
The Travis CI passes all the test cases, but it doesn't test the `--enable-flow-api`, so it doesn't check exactly this change: https://travis-ci.org/github/MassimoGirondi/fastclick/builds/738497022

tbarbette:
Great, thanks!

PR #280, opened 2020-11-16
Fixed DPDK configuration

gkatsikas:
Covers both meson and make build systems


tbarbette:
Seems good, thanks!

PR #281, opened 2020-11-16
Support for DPDK v20.11

gkatsikas:
Covers:
 * DPDK v20.11-rc0
 * DPDK v20.11-rc1
 * DPDK v20.11-rc2
 * DPDK v20.11-rc3
 * DPDK v20.11-rc4

It looks like that the PKT_RX_TIMESTAMP is deprecated.


tbarbette:
It's now a dynamic field. We should refuse to parse at least if it's not supported, but the best would be to use the dynamic field if version is 20.11 or higher.

tbarbette:
Ok I added the support for dynamic field in https://github.com/tbarbette/fastclick/tree/pr-dpdk-20.11, so I wait for travis because I did not test the old way and then merge

PR #293, opened 2020-12-22
Add FromMinDump and ToMinDump elements

MassimoGirondi:
Similarly to *IPSummaryDump elements, record and generation of the packets is
done with a minimal size trace containing only the 5-tuple of each
packet.


tbarbette:
Thanks! The problem is due to the current master.

PR #294, opened 2021-01-03
Fix testing

tbarbette:


tbarbette:
Pushed upstream as errors were only coverage (fixed while pushing)

PR #312, opened 2021-02-11
PacketBatch: Remove inheritance to make the batching method abstract

tbarbette:
Towards allowing batches to be either vector, LL or whatever we have to
stop casting PacketBatch to the first packet, as vectors would be
independent objects


aliireza:
This is a nice change. 
The idea is to implement vector batches soon, right?

tbarbette:
Yes, it's something I wanted to try since a long time :) Also if we implement a vectorized "only" version of XCHG in DPDK it makes more sense.

PR #313, opened 2021-02-11
integrating LLVM passes to packetmill.cc

aliireza:
@tbarbette Can you check this?

I still need to add some checks, but this is the basic version. I could still do the click-llvm-opt or separate some parts if you think it is necessary.

### DIFF
--- a/DISTFILES
+++ b/DISTFILES
@@ -432,6 +432,7 @@ lexer.hh
 libdivide.h
 list.hh
 llrpc.h
+llvmutils.hh
 machine.hh
 master.hh
 md5.h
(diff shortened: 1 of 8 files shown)


tbarbette:
Can you move the functions that do the reordering etc in something like lib/ or tools/lib ? So we keep packetmill.cc just as a "main".
This morning I had plan to fix the CI issues also... :p

aliireza:
Yes! I wasn't sure where to put it :-p 

Nice!

aliireza:
@tbarbette Can you check now? :-p 

Should we wait for the tests before merging?

tbarbette:
Can you rebase now I fixed the CI? :p

aliireza:
Passed :-p

PR #316, opened 2021-02-19
FromDPDKDevice: add unset option to PAUSE

MassimoGirondi:
unset option for PAUSE was specified in the documentation but not
accepted by the parser


tbarbette:
Well it's usually the documentation lagging on the features :p That's a first. Thanks! :p

PR #317, opened 2021-02-21
Disassemble/Inline AllAnno Struct

aliireza:
There is an issue with the reordering pass when we use -O3, but I will fix it soon.

### DIFF
--- a/configure
+++ b/configure
@@ -832,6 +832,7 @@ enable_dpdk_inside
 enable_dpdk_packet
 enable_dpdk_softqueue
 enable_pool_inlining
+enable_inlined_allanno
 enable_flow_api
 enable_zerocopy
 enable_clone
@@ -1574,6 +1575,9 @@ Optional Features:
                           Do not optimize pool_prepare_data_burst function
                           before data structure reordering done at the IR
                           level.
+    --enable-inlined-allanno
+                          Inline AllAnno struct into Packet class,
+                          facilitating the data structure reordering.
       --enable-flow-api   Support for the HW flow classification API. Also
                           needs DPDK>=20.02
     --disable-zerocopy    disable Zero Copy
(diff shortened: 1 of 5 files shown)


aliireza:
I think it is fixed now.

configure
```diff
@@ -1574,6 +1575,9 @@ Optional Features:
                           Do not optimize pool_prepare_data_burst function
                           before data structure reordering done at the IR
                           level.
+    --enable-inlined-allanno
```
tbarbette, line 1578: Should add two spaces here, no?

include/click/packet.hh
```diff
@@ -837,6 +842,10 @@ class Packet { public:
 #if !CLICK_PACKET_USE_DPDK
 	char timestamp[sizeof(Timestamp)];
 #endif
+#if INLINED_ALLANNO
```
tbarbette, line 845: Why do we have to change the name?

include/click/packet.hh
```diff
@@ -1265,6 +1279,8 @@ Packet::next() const
     return (Packet *)(skb()->next);
 #elif CLICK_PACKET_USE_DPDK
     return all_anno()->next;
+#elif INLINED_ALLANNO
+    return nextp;
```
tbarbette, line 1283: Isn't there a way to have a macro to prevent all thse ifdef? Like _aa is a typedef/#define for this? Or change "_aa." per something likge getanno() which returns this or Anno* according to the mode?

tbarbette (changes requested):

configure
aliireza: I just ran autoconf, I will fix it.

include/click/packet.hh
aliireza: Compile error! :-p It seems we cannot have an inline function and a variable with the same name.

include/click/packet.hh
aliireza: This is actually a nice idea! packet.hh is too messy. I will try what you suggested. Thanks :)

tbarbette (approved):

PR #323, opened 2021-03-09
Fix compilation on non x86

MassimoGirondi:
- When checking for AVX2 support on non x86 platform the control is skipped, otherwise it fails the configuration. Should this control be implemented differently?
- Fix a wrong compilation flag guard regarding click_get_cycles


tbarbette:
Seems good, thanks!

Issue #320, opened 2021-02-26
Problem with XL710 Intel NIC


sarsanaee:
Hey, 

I'd like to test fastclick with `XL710 Intel NIC`. However, something seems to be wrong. It actually works properly with `Mellanox ConnectX-4 Lx`.

Here is the error, I'm getting:
```
conf/dpdk/dpdk-bounce.click:11: While initializing ‘FromDPDKDevice@1 :: FromDPDKDevice’:
  The number of receive descriptors is 100 but needs to be between 0 and 0
Router could not be initialized!
```

I don't understand why it expects 0 for the receive descriptors! It is just the simplest script,`dpdk-bounce.click`.

Thanks

sarsanaee:
I believe the issue is coming from here.

https://github.com/tbarbette/fastclick/blob/69d731ec18ecb8d3b5388719f6d143dffa5c6958/lib/dpdkdevice.cc#L594

We are expecting the `rte_eth_dev_info_get` to fill `dev_info` no? Is it a driver issue then?

Thanks,
Alireza

tbarbette:
Hi!
Yes it seems there's actually no guarantee these fields are filled by the driver. I'll add a condition.
It's weird the driver does not set it though... I did use XL710 and did not have that problem. Are you using VFs or something specific about the driver? Which DPDK version?
Tom

sarsanaee:
Hi Tom,

Thank you for the reply.

No, I don't use VFs.

I loaded igb_uio downloaded from kmod repository. DPDK version is 20.10.

Thanks,
Alireza

magaboomchen:
Hi!
I meet a similar problem with XL710.
I just run a simple L2-fwd click using DPDK. Then FastClick reports:
EAL: Probing VFIO support...
EAL: VFIO support initialized
EAL: PCI device 0000:05:02.0 on NUMA socket 0
EAL:   probe driver: 8086:154c net_i40e_vf
Initializing DPDK
i40evf_dev_configure(): VF can't disable HW CRC Strip
/home/user/l2fwd.click:2: While initializing ‘out0 :: ToDPDKDevice’:
  Cannot initialize DPDK port 0 with 1 RX and 1 TX queues
  Error -22 : Unknown error -22
Router could not be initialized!

Here is my click conf:
in0 :: FromDPDKDevice(0);
out0 :: ToDPDKDevice(0);
in0 -> EtherMirror() -> out0;

And here is the driver of XL710:
driver: i40e
[...]

PR #326, opened 2021-03-15
Fix compilation with old DPDKs

tbarbette:


elements/tcpudp/shifter.cc
```diff
@@ -66,18 +66,19 @@ inline void Shifter::process(Packet *p) {
         q->rewrite_ipport(htonl(ip_src), htons(sport), 0, ip_proto == IP_PROTO_TCP);
         q->rewrite_ipport(htonl(ip_dst), htons(dport), 1, ip_proto == IP_PROTO_TCP);
     }
+    return q;
```
tbarbette, line 69: @MassimoGirondi there were a few problems here. uniqueify may return a new packet, so you have to return it and use that one, you can't use the old one afterwards

elements/tcpudp/shifter.cc
```diff
@@ -66,18 +66,19 @@ inline void Shifter::process(Packet *p) {
...
 }
 
 #if HAVE_BATCH
 void Shifter::push_batch(int, PacketBatch *batch) {
-    FOR_EACH_PACKET(batch, p) { process(p); }
+    EXECUTE_FOR_EACH_PACKET(process,batch);
```
tbarbette, line 74: @MassimoGirondi this would replace the packet in the batch if it changed

elements/tcpudp/shifter.cc
```diff
@@ -66,18 +66,19 @@ inline void Shifter::process(Packet *p) {
...
     output_push_batch(0, batch);
 }
 #endif
 
 Packet *Shifter::simple_action(Packet *p) {
-    process(p);
-    output_push(0, p);
```
tbarbette, line 80: Also simple_action is a special function, you can't push in that as the system will push the packet returned by the function. So here you would actually push 2 packets out for every packet in, leaking.

tbarbette:
Will also fix stuffs in #320

PR #330, opened 2021-04-02
Merge most of PacketMill

tbarbette:
Just not the RAND_ALIGN which was only a paper feature, and not
DPDK_INSIDE that was to simulate VPP model, not entierly finished and is
not the right choice anyway.


aliireza:
Looks good! 
Do you want me to run some tests before merging? :-p

tbarbette:
If you have time to re-run a packetmill vs fastclick test, it would be great

aliireza:
We are good! 😉

tbarbette:
Great! Thanks for checking :)

PR #332, opened 2021-04-16
New element: SetTCPChecksumIP6 - Calculates TCP/IPv6 checksum and sets the TCP header field.

kkovacs:
The existing `SetTCPChecksum` element only supports TCP over IPv4. This is a new element that does the same for TCP over IPv6.

The new element's naming convention follows `FastUDPSource` and `FastUDPSourceIP6`, and is also the reason I've put this in `tcpudp/` instead of `ip6/`.

The code is based mostly on the original `SetTCPChecksum` and the way `FastUDPSourceIP6` calculates checksum for its packets.

### DIFF
--- a/elements/tcpudp/settcpchecksumip6.cc
+++ b/elements/tcpudp/settcpchecksumip6.cc
@@ -1,5 +1,5 @@
 /*
- * SetTCPChecksumIP6.{cc,hh} -- sets the TCP header checksum
+ * settcpchecksumip6.{cc,hh} -- sets the TCP header checksum for IPv6 packets
  * Robert Morris
  *
  * Copyright (c) 1999-2000 Massachusetts Institute of Technology
--- a/elements/tcpudp/settcpchecksumip6.cc
+++ b/elements/tcpudp/settcpchecksumip6.cc
@@ -22,6 +22,7 @@
 #include <click/error.hh>
 #include <clicknet/ip6.h>
 #include <clicknet/tcp.h>
+
 CLICK_DECLS
 
 SetTCPChecksumIP6::SetTCPChecksumIP6()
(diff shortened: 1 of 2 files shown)


kkovacs:
I do see that a "CodeQL" build step is failing on the PR (but all builds are successful), but this one does not seem related to my code, it says:

> _"Semmle autobuild: no supported build system detected. We were unable to automatically build your code. Please replace the call to the autobuild action with your custom build steps."_

elements/tcpudp/settcpchecksumip6.cc
```diff
@@ -0,0 +1,71 @@
+/*
+ * settcpchecksumip6.{cc,hh} -- sets the TCP header checksum for IPv6 packets
+ * Robert Morris
```
tbarbette, line 3: Can you add your name and copyright institution? Maybe add "based on settcpchecksum.{cc,hh}"

tbarbette (approved):
Thank you very much! Could you just set the copyright correctly?

tbarbette:
Yes the codeql seemed to be a false positive, I'll check after it re-ran.

elements/tcpudp/settcpchecksumip6.cc
kkovacs: Yes, of course. I didn't want to claim too much credit, but the way you suggest it is a good way, referencing the others. Pushed!

kkovacs:
Copyright was changed as asked (thanks!), now I think only CodeQL is acting up again.

tbarbette:
I found the problem, codeql does not compile with --enable-ip6 and therefore the element is missing a symbol. It's actually legit Should be fixed with the last commit. I'll merge when it passes. Thanks!

tbarbette:
Bad luck... Github has an issue with name resolution right now and cannot resolve fast.dpdk.org... Will re-launch the CI in a bit.

kkovacs:
Thanks for the quick merge, Tom! And thank you for doing fastclick!

PR #350, opened 2021-11-18
Created  toexternalprocess.cc

regufo:
The Element ToExternalProcess is a not batch element that push data in a shared memory initialized by another process who lives out of FastClick. The referiment to the shared memory must be passed as a parameter. This Element consider the memory as diveided in virtual pages, each times it fill a virtual page it sets a shared semaphore.
This class could be used to implements cybersecurity tools as WAF, NAC, IPS, IDS, or tools who neesd to interact with external data and functions.
When the memory is full, this element go back to the first memory byte and rewrites the old contens. The external process should works the content before overwriting occurs.
The parameters are:

- SHARED_MEMORY_ID : positional mandatory parameter, it is the id of shared memory initialized by the external process
- SEMAPHORE_ID: positional mandatory parameter, it is the id a shared semaphore who will be set by the the element when the virtual page will be filled
- BUFFER_SIZE: positional parameter, it is an integer who specify how many rows the buffer contains (a row is a memory space sufficient to write a package, defined in the PACKAGE_SIZE parameter)
- NUMBER_OF_PAGES: positional parameter, it is an integer who specifies how many pages the buffer is divided into
- ENTEER_PACKET : positional boolean parameter, if it is "false" the Element will copy the first 32 bytes in the shared memory, else all data ll be copied.
- PACKAGE_SIZE :  positional parameter, it is the size of a packet, if the Element will receive Ethernet frames this should take the value 1522

Configuration Example:
tep :: ToExternalProcess(SHARED_MEMORY_ID 123, SEMAPHORE_ID 255, BUFFER_SIZE 1024, NUMBER_OF_PAGES 128, ENTEER_PACKET 1, PACKAGE_SIZE 1522);

### DIFF
--- a/elements/standard/toexternalprocess.cc
+++ b/elements/standard/toexternalprocess.cc
@@ -1,0 +1,169 @@
+#include <click/config.h>
+#include <click/error.hh>
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/shm.h>
+#include <sys/sem.h>
+#include <clicknet/ip.h>
+#include <clicknet/ether.h>
+
+#include "toexternalprocess.hh"
+CLICK_DECLS
+
+union semun
+{
+    int val;
+    struct semid_ds *buf;
+    ushort *array;
+} semaphor_state;
+
+ToExternalProcess::ToExternalProcess()
+{
+    empty_data = (char *)malloc(sizeof(char) * _package_size);
+    for (int i = 0; i < _package_size; i++)
+    {
+        empty_data[i] = '\0';
+    }
+}
+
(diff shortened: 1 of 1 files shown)


tbarbette:
Hi @regufo !

First could you send both files in one pull request? You can update this one by pushing to the same branch :)

Do you have an example of how a "client" of this element could be used? A library that allows to read the memory from any software?

Do you have some comparison to other techniques like passing packets to a socket? What would be the interest compared to To/FromDPDKRing?

Also the "userlevel" directory would be better suited, as this element will not work in Kernel.

Thanks,
Tom

PR #351, opened 2021-11-18
NICBench fixes

MassimoGirondi:
FlowRuleInstaller is now always compiled, but it will support the update action only when rte_flow_update is found in the DPDK rte_flow library.

And some other small dangling commits...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ compile_commands.json
 *.resolution.txt
 *.lto.o
 build/*
+userlevel/packetmill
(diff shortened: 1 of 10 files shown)


tbarbette:
Ah yes, thanks ! How did it happen that the CI worked but rte_flow_update is not there in the normal DPDK...?

MassimoGirondi:
Actually, it was not compiling FlowRuleInstaller due to the missing Json provision. So there was no difference between normal and patched versions for FastClick.

PR #365, opened 2022-07-18
update DPDK arguments when no huge pages are used

gkatsikas:
Signed-off-by: Georgios Katsikas <gkatsikas@ubitech.eu>


tbarbette:
👍

gkatsikas:
It was not a problem of -v, maybe the debug verbosity is causing the errors

PR #372, opened 2022-11-28
Merge new connection tracking element

tbarbette:
Following the paper High-speed Connection Tracking in modern servers lead by @MassimoGirondi  (https://girondi.net/pub/conntrack21.pdf) here are finally a few changes merged in FastClick. It's nearly a whole rewrite to remove all the possibilities and layers we had to compare performance with various angles and have a more minimal ancestor, VirtualFlowManagerIMP.

FlowIPManagerIMP is now an alias to FlowIPManager_CuckooPP, a per-thread Cuckoo++ hashtable from https://arxiv.org/pdf/1712.09624.pdf . It is duplicated per-core, and each core has its own timerwheel.

The old FlowIPManagerIMP becomes FlowIPManager_DPDK. It now supports timerwheel too, per core also.

The FlowIPManagerMP stays the DPDK LF cuckoo hash-table. That versions is lock-free internally, to be used when the cores absolutely need to share state.

Generally all those element do not need FlowControlBlock usage counting as they rely on timers. So --enable-flow-dynamic is now optional. It's still needed for --enable-ctx, as the context subsystem (middleclick) needs this.

I did a few fixes and removed a few warnings at the same time.


tbarbette:
image
Quick graph (also to verify it works) with CAIDA.

PR #375, opened 2023-01-11
[Fix build without DPDK] Add DPDK dependency to DeviceBalancer

louisna:
Add DPDK dependency to DeviceBalancer because it requires `rte_flow.h` which is only available with DPDK. This modification repairs the build of fastclick without DPDK.


tbarbette:
Actually it shouldn't. This element can also work with XDP programs in normal Kernels. But I'll accept the fix meanwhile (when the CI passes). Thanks !

tbarbette:
Forgot to merge this, sorry.

PR #376, opened 2023-01-27
push_batch not properly ifdef'd, failing to build if batch is disabled

ekline2:
Fixed by ifdefing in header


tbarbette:
Thanks!

PR #377, opened 2023-02-07
Updated flags to be prefixed by RTE_ for DPDK 22

ThornyRufus:
Adapted the flags to be compatible with DPDK 22 as stated there : https://mails.dpdk.org/archives/dev/2021-July/215841.html


elements/analysis/aggcountervector.hh
```diff
@@ -133,7 +133,7 @@ AggregateCounterVector::find_node(uint32_t a, const Packet* p, bool &outdated)
         struct rte_mbuf* p_mbuf;
         p_mbuf = (struct rte_mbuf *) p->destructor_argument();
 
-        if (!(p_mbuf->ol_flags & PKT_RX_FDIR_ID))  {
+        if (!(p_mbuf->ol_flags & RTE_MBUF_F_RX_FDIR_ID))  {
```
tbarbette, line 136: This will break the compilation with any older DPDK, we want to keep compatibility. Please enclose all changes in an ifdef as I showed you on teams, or globally define an alias.

tbarbette (changes requested):

include/click/tcphelper.hh
```diff
@@ -24,6 +24,13 @@
 #include <rte_ip.h>
 #endif
 
+#if RTE_VERSION < RTE_VERSION_NUM(22,07,0,0)
```
tbarbette, line 27: Monte le dans le ifdef HAVE_DPDK pour fixer l'erreur de compil.

tbarbette:
Thanks!

PR #378, opened 2023-03-24
feat: support uint64 token bucket rates

gkatsikas:

PR #395, opened 2024-08-21
Support 64-bit bandwidths

mihaibrodschi:
Based on #378 by gkatsikas.

Updates TokenBucket, GapRate and bw argument parsing.

### DIFF
--- a/elements/userlevel/jiffieclock.cc
+++ b/elements/userlevel/jiffieclock.cc
@@ -61,13 +61,13 @@ JiffieClock::initialize(ErrorHandler*) {
 
 void JiffieClock::run_timer(Timer* t) {
     Timestamp current_time = Timestamp::now_steady();
-    int64_t delta = (current_time - last_jiffies_update).msec() / (1000 / CLICK_HZ);
+    int64_t delta = (current_time - last_jiffies_update).usec() / (1000000 / CLICK_HZ);
     if (delta > 0) {
         if (unlikely(delta > _minprecision)) {//Accept a little jump from time to time, but not double jump
             //We try all the click threads
             int nt = (t->home_thread_id() + 1) % master()->nthreads();
             if (nt == home_thread_id()) {
-                click_chatter("Click tasks are too heavy and the jiffie accumulator cannot run at least once every %dmsec, the user jiffie clock is deactivated.",_minprecision);
+                click_chatter("Click tasks are too heavy and the jiffie accumulator cannot run at least once every %d jiffies, the user jiffie clock is deactivated.",_minprecision);
                 click_jiffies_fct = &click_timestamp_jiffies;
                 return;
             }
@@ -78,7 +78,7 @@ void JiffieClock::run_timer(Timer* t) {
         jiffies += delta;
         last_jiffies_update = current_time;
     }
-    t->schedule_at_steady(last_jiffies_update + Timestamp::make_msec(1000 / CLICK_HZ));
+    t->schedule_at_steady(last_jiffies_update + Timestamp::make_usec(1000000 / CLICK_HZ));
 }
 
 click_jiffies_t JiffieClock::read_jiffies(void* data) {
@@ -94,7 +94,7 @@ bool JiffieClock::run_task(Task*) {
     //TODO : this should be rcu protected
     click_jiffies_fct_data = this;
     click_jiffies_fct = &read_jiffies;
-    _timer.schedule_after_msec(1000 / CLICK_HZ);
+    _timer.schedule_after(Timestamp::make_usec(1000000 / CLICK_HZ));
     return true;
 }
 
--- a/elements/userlevel/jiffieclock.cc
+++ b/elements/userlevel/jiffieclock.cc
@@ -61,13 +61,13 @@ JiffieClock::initialize(ErrorHandler*) {
 
 void JiffieClock::run_timer(Timer* t) {
     Timestamp current_time = Timestamp::now_steady();
-    int64_t delta = (current_time - last_jiffies_update).usec() / (1000000 / CLICK_HZ);
+    int64_t delta = (current_time - last_jiffies_update).msec() / (1000 / CLICK_HZ);
     if (delta > 0) {
         if (unlikely(delta > _minprecision)) {//Accept a little jump from time to time, but not double jump
             //We try all the click threads
             int nt = (t->home_thread_id() + 1) % master()->nthreads();
             if (nt == home_thread_id()) {
-                click_chatter("Click tasks are too heavy and the jiffie accumulator cannot run at least once every %d jiffies, the user jiffie clock is deactivated.",_minprecision);
+                click_chatter("Click tasks are too heavy and the jiffie accumulator cannot run at least once every %dmsec, the user jiffie clock is deactivated.",_minprecision);
                 click_jiffies_fct = &click_timestamp_jiffies;
                 return;
             }
@@ -78,7 +78,7 @@ void JiffieClock::run_timer(Timer* t) {
         jiffies += delta;
         last_jiffies_update = current_time;
     }
-    t->schedule_at_steady(last_jiffies_update + Timestamp::make_usec(1000000 / CLICK_HZ));
+    t->schedule_at_steady(last_jiffies_update + Timestamp::make_msec(1000 / CLICK_HZ));
 }
 
 click_jiffies_t JiffieClock::read_jiffies(void* data) {
@@ -94,7 +94,7 @@ bool JiffieClock::run_task(Task*) {
     //TODO : this should be rcu protected
     click_jiffies_fct_data = this;
     click_jiffies_fct = &read_jiffies;
-    _timer.schedule_after(Timestamp::make_usec(1000000 / CLICK_HZ));
+    _timer.schedule_after_msec(1000 / CLICK_HZ);
     return true;
 }
 
--- a/include/click/glue.hh
+++ b/include/click/glue.hh
@@ -528,10 +528,10 @@ typedef click_jiffies_t (*click_jiffies_fct_t)(void*);
 extern click_jiffies_fct_t click_jiffies_fct;
 extern void* click_jiffies_fct_data;
 #  define click_jiffies()      (click_jiffies_fct(click_jiffies_fct_data))
-#  define CLICK_HZ           1000
+#  define CLICK_HZ           1000000
 # else
 #  define click_jiffies()      (click_timestamp_jiffies(0))
-#  define CLICK_HZ           1000
+#  define CLICK_HZ           1000000
 # endif
 # define HAS_LONG_CLICK_JIFFIES_T  1
 # define click_jiffies_less(a, b)	((click_jiffies_difference_t) ((a) - (b)) < 0)
--- a/include/click/glue.hh
+++ b/include/click/glue.hh
@@ -528,10 +528,10 @@ typedef click_jiffies_t (*click_jiffies_fct_t)(void*);
 extern click_jiffies_fct_t click_jiffies_fct;
 extern void* click_jiffies_fct_data;
 #  define click_jiffies()      (click_jiffies_fct(click_jiffies_fct_data))
-#  define CLICK_HZ           1000000
+#  define CLICK_HZ           1000
 # else
 #  define click_jiffies()      (click_timestamp_jiffies(0))
-#  define CLICK_HZ           1000000
+#  define CLICK_HZ           1000
 # endif
 # define HAS_LONG_CLICK_JIFFIES_T  1
 # define click_jiffies_less(a, b)	((click_jiffies_difference_t) ((a) - (b)) < 0)
--- a/elements/ip/iprewriterbase.hh
+++ b/elements/ip/iprewriterbase.hh
@@ -429,7 +429,7 @@ IPRewriterBase::search_migrate_entry(const IPFlowID &flowid, per_thread<T> &vsta
 {
     //If the flow does not exist, it may be in other thread's stack if there was a migration
     if (vstate->rebalance > 0 &&
-			click_jiffies() - vstate->rebalance < (uint64_t) THREAD_MIGRATION_TIMEOUT * CLICK_HZ / MS_PER_S) {
+			click_jiffies() - vstate->rebalance < (uint64_t) THREAD_MIGRATION_TIMEOUT * CLICK_HZ / 1000) {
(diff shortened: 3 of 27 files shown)


tbarbette:
Thanks! Can you take a look at the failing tests? I can help you if you don't get them.

mihaibrodschi:
Looks like I used a symbol defined by DPDK, so non-DPDK builds are failing.

tbarbette:
Can you add an ifundef then?  Probably in glue.hh if used multiple times, or just where it is used if just once.

tbarbette:
That one looks functional. You can use make check locally to debug. I saw some MS value changed to US, maybe that's the problem. You should try to keep configuration compatible to previous behavior.

mihaibrodschi:
Some BigInt tests are broken. They were written to compare 64-bit multiplication and division with the same operation using 2-limb BigInts, where each limb was 32-bit. Now (with 64-bit limbs) to test the same way we would need to compare with 128-bit operations, which aren't necessarily supported on all platforms.

tbarbette:
Ah ok I understand. If bigint was defined as 8 bytes as in some languages then it would probably be best to create a Bigint128 or so instead of changing Bigint as some part of the code might assume some behavior that would be hard to catch in test (as in wraparound). What do you think?

tbarbette:
Can you test with shift and masks? Or propose any other resolution that is acceptable :)

tbarbette:
Do you actually need bigint? We might keep this problem/discussion separate.

mihaibrodschi:
BigInt is used by TokenBucket in a few places. Not sure if replacing it is the best solution. The problem is the test code.

tbarbette:
Yes, tests being broken is why I never merged. Can you try to reverse the testing to 32bits, and add maybe a few minimal 64bit ones? At least invoke it to verify it compiles?

include/click/glue.hh
```diff
@@ -528,10 +528,10 @@ typedef click_jiffies_t (*click_jiffies_fct_t)(void*);
 extern click_jiffies_fct_t click_jiffies_fct;
 extern void* click_jiffies_fct_data;
 #  define click_jiffies()      (click_jiffies_fct(click_jiffies_fct_data))
-#  define CLICK_HZ           1000
+#  define CLICK_HZ           1000000
```
tbarbette, line 531: is it really necessary to change CLICK_HZ ? It seems like a major change that could break things. There might be places and external elements which assumes (it's bad but it could happen) that CLICK_HZ is 1000. In the Kernel mode it's even just the kernel timer ticks.
Timer themselves do not use jiffies. It's probably only used for simplification somewhere.

elements/userlevel/jiffieclock.cc
```diff
@@ -61,13 +61,13 @@ JiffieClock::initialize(ErrorHandler*) {
 
 void JiffieClock::run_timer(Timer* t) {
     Timestamp current_time = Timestamp::now_steady();
-    int64_t delta = (current_time - last_jiffies_update).msec() / (1000 / CLICK_HZ);
+    int64_t delta = (current_time - last_jiffies_update).usec() / (1000000 / CLICK_HZ);
```
tbarbette, line 64: Are you using JiffieClock or did you catch this would break?

mihaibrodschi: I just noticed it.

include/click/glue.hh
mihaibrodschi: The reason I changed it is that BandwidthRatedUnqueue puts the task to sleep when its TokenBucket is empty, and schedules a timer to wake it up once enough tokens have accumulated. If the timer resolution isn't high enough, it will sleep for too long, and cause huge latency spikes.
See [elements/standard/bwratedunqueue.cc:88](https://github.com/tbarbette/fastclick/blob/main/elements/standard/bwratedunqueue.cc#L88).

mihaibrodschi: Alternatively, I could implement a different helper class for TokenBucket to use microseconds/nanoseconds instead of jiffies.

tbarbette: Yes I think instead of make_jiffies you can simply use make_usec

mihaibrodschi:
The [RatedUnqueue test](https://github.com/tbarbette/fastclick/blob/main/test/standard/RatedUnqueue-01.clicktest) fails because it expects an inaccurate result (909|910), but receives the accurate one (1000). Is it OK to update the test file?

tbarbette:
Weird.. I'll take a look.

tbarbette:
Seems good now, thanks ! Sorry for the delay.

mihaibrodschi:
The latest PR merged (#398) undid some of the changes made in this one.

tbarbette:
@mihaibrodschi there are two failures in the advanced test (see https://forge.uclouvain.be/ensg/fastclick/-/pipelines/53850). Not sure if you have access, but basically, I am building with 
```./configure CXXFLAGS="-std=gnu++11" --disable-batch --enable-simtime && make clean && make -j 16 && make check```

and then I get:

```
./test/standard/BandwidthRatedSplitter-01.clicktest:41: standard error has unexpected value starting at line 3
./test/standard/BandwidthRatedSplitter-01.clicktest:41: stderr:3: expected '20'
./test/standard/BandwidthRatedSplitter-01.clicktest:41: stderr:3: but got  '21'
./test/standard/BandwidthRatedUnqueue-01.clicktest:39: standard error has unexpected value starting at line 5
./test/standard/BandwidthRatedUnqueue-01.clicktest:39: stderr:5: expected '71'
./test/standard/BandwidthRatedUnqueue-01.clicktest:39: stderr:5: but got  '70'
```

I would say it's because you added a better precision but it does not happen in all cases. Any idea?

**Comment by mihaibrodschi** [comment]
@tbarbette I think the results are correct.
For the first test, the setup is:
```
BandwidthRatedSpliter-01.clicktest:
InfiniteSource(LENGTH 100)
	-> Queue(10)
	-> u1 :: BandwidthRatedUnqueue(RATE 2000Bps)
	-> s1 :: BandwidthRatedSplitter(RATE 200Bps)
	-> c1 :: Counter
	-> Discard;
Script(wait 10, read c1.count, read c2.count, read c3.count, read c4.count, write stop);
Expected 20, got 21.
```

The BandwidthRatedSplitter uses a token bucket, which has an initial amount of tokens equal to tb_bandwidth_thresh = 131072 ([see elements/standard/ratedunqueue.cc:77](https://github.com/tbarbette/fastclick/blob/acdfb021699b169e3e837545bc086c772dec086c/elements/standard/ratedunqueue.cc#L77)).
To check if a packet can go through, the RatedSplitter compares the token bucket's contents to the tb_bandwidth_thresh. Since there are enough tokens in the bucket at the beginning, one packet goes through immediately.
Over the next 10 seconds, another 20 packets go through at the specified rate, so the total is 21.

For the second test, the setup is:
```
BandwidthRatedUnqueue-01.clicktest:
InfiniteSource(LENGTH 100)
	-> Queue(10)
	-> u2 :: BandwidthRatedUnqueue(RATE 200Bps, BURST_BYTES 5000)
	-> c2 :: Counter
	-> Discard;
Script(wait 10, read c1.count, read c2.count, read c3.count, read c4.count, write stop);
Expected 71, got 70.
```
In this case, the BWRatedUnqueue has a token bucket with an initial token count of 5000 + tb_bandwidth_thresh.
Thus, 51 packets are immediately allowed through. After this, the BWRatedUnqueue task only wakes up after enough time has passed to replenish its tokens back to tb_bandwidth_thresh ([see elements/standard/bwratedunqueue.cc:88](https://github.com/tbarbette/fastclick/blob/main/elements/standard/bwratedunqueue.cc#L88)).
In this case, the problem seems to be a slight inaccuracy in the timing. In 10 seconds, the task should wake up 20 times, but it only wakes up 19 times. If I extend the time by 2 microseconds:
```Script(wait 10000002us, read c1.count, read c2.count, read c3.count, read c4.count, write stop);```
the test passes with the expected result of 71.
I also noticed that BandwidthRatedSplitter-01.clicktest contains a test case very similar to this one:
```
InfiniteSource(LENGTH 100)
	-> Queue(10)
	-> u2 :: BandwidthRatedUnqueue(RATE 2000Bps)
	-> s2 :: BandwidthRatedSplitter(RATE 200Bps, BURST_BYTES 5000)
	-> c2 :: Counter
	-> Discard;
```
with an expected result of 70.

**Comment by tbarbette** [comment]
So there was too problems, first with batching enabled, the default burst is still to read 32 packets. So reading 32 packets leads to very large imprecision in terms of token overcommitment.

With a BURST of 1, all tests are now failing for this 20/21 reason.

The timing before your patch if this:
```
1000000000.000000008:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000000.530000076:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000001.030000138:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000001.530000200:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000002.030000262:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000002.530000324:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000003.030000386:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000003.530000448:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000004.030000510:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000004.530000572:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000005.030000634:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000005.530000696:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000006.030000758:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000006.530000820:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000007.030000882:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000007.530000944:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000008.030001006:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000008.530001068:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000009.030001130:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000009.530001192:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
c1.count:
20
```

After your patch it is : 
```
1000000000.000000008:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000000.480001070:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000000.980001132:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000001.480001194:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000001.980001256:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000002.480001318:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000002.980001380:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000003.480001442:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000003.980001504:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000004.480001566:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000004.980001628:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000005.480001690:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000005.980001752:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000006.480001814:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000006.980001876:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000007.480001938:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000007.980002:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000008.480001062:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000008.980001124:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000009.480001186:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
1000000009.980001248:  100 | 52616e64 6f6d2062 756c6c73 68697420 696e2061 20706163
c1.count:
21
```
We see the bucket is refilled a bit sooner, so 3 packets go up in the first second.

This is only happening with ``--simtime``. With real time it does not happen.

**Comment by mihaibrodschi** [comment]
> So there was another problem. First with batching enabled, the default burst is still to read 32 packets. So reading 32 packets leads to very large imprecision in terms of token overcommitment.

This is because of [this code](https://github.com/tbarbette/fastclick/blob/887a6db8a6af80ac85dbc3c04b7b64dbd0724898/elements/standard/bwratedunqueue.cc#L55) in bwratedunqueue.cc, correct?
```
    if (_tb.contains(tb_bandwidth_thresh)) {
#if HAVE_BATCH
        if (in_batch_mode) {
            PacketBatch* batch = input(0).pull_batch(_burst);
```
It does not check if there are enough tokens for the entire batch, only for one packet.
This is an older bug, isn't it?

As for the slight timing difference, is it an incorrect result? If so, I'll see what I can do to fix it.

mihaibrodschi:
Perhaps a cleaner solution for batching in BWRatedUnqueue would be to store the packets which can't be pushed due to insufficient tokens in a small queue (of size _burst).

tbarbette:
Yes indeed. It's not really a bug, it's a different "default" behavior (which is not appreciable).
I thought of the "internal queue" solution but maybe that should be another element on top of BWUnqueue because checking the queue might impact a bit the performance.
I pushed the update on the PRfixbw branch. The behavior is now constant at least.
Now the right behavior depends on what to expect for the default amount of tokens at t=0...

PR #396, opened 2024-09-18
Merging code from HIRT

tbarbette:
Changes improving IPv6 and adding SRv6 encap/decap and some processing capabilities to showcase the HIRT paper to be presented at ICNP'24.

It also adds the library elements fo network-layer FEC used in the aforementioned paper.

@louisna was mostly involved in the code, as well as me.

Before merging, I need to make the FEC code an argument to ./configure because there's probably some dependencies we don't want for a default build.


tbarbette:
Will need to redo the PR to squash

PR #398, opened 2024-10-01
Merge most code from HIRT (ICNP'24)

tbarbette:
Changes improving IPv6 and adding SRv6 encap/decap and some processing capabilities to showcase the HIRT paper to be presented at ICNP'24.

It also adds the library elements fo network-layer FEC used in the aforementioned paper. However the FEC elements themselves use RLC code that is under patent, so if you want them please ask Louis. For the same reasons, I had to squash all commits as once, sorry about the huge change.

Improvements:
 - updated gitlab CI with much more tests
 - SRV6 encap and decap elements
 - Fix compatibility problem with aarch64
 - Fix compatibility issues with MacOS

@louisna

 was mostly involved in the code, and then me.

Co-authored-by: default avatarLouis Navarre <louis.navarre@uclouvain.be>


tbarbette:
Finally I think this one is ready...

PR #402, opened 2024-10-24
Fixes for 64-bit rates

mihaibrodschi:
This adds some uint32->uint64 conversions which I missed in #395.
They affect RatedUnqueue, RatedSource and BandwidthMeter.


tbarbette:
Thanks!

PR #403, opened 2024-10-28
Set BURST to 1 by default for bandwidthratedunqueue

tbarbette:
Reading 32 packets will lead to very large imprecision in term of byts

### DIFF
--- a/elements/standard/ratedunqueue.cc
+++ b/elements/standard/ratedunqueue.cc
@@ -38,7 +38,7 @@ int
 RatedUnqueue::configure(Vector<String> &conf, ErrorHandler *errh)
 {
     if (Args(this, errh).bind(conf)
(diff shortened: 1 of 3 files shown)


tbarbette:
@mihaibrodschi  this looks good now. With the fix it's consistent and I think the new values make sense.

mihaibrodschi:
Great! Just one question, from a performance standpoint, isn't setting the burst to 1 the same as (or worse than) disabling batching?

tbarbette:
Yes, but the user can set BURST 32, and anyway one should avoid pull mode. In push mode the problem does not happen.

tbarbette:
I will add a warning.

tbarbette:
Just pushed. If you're up to it you can create something like ExactBandwidthRatedUnqueue which has this internal queue :)

PR #410, opened 2025-02-24
Fix compilation with --enable-stats=2

### LINKED ISSUE #409: Failed compilation with --enable-stats=2
When configuring FastClick with --enable-statistics=2, the compilation fails. This does not happen with =1 or =0.

It seems to be due to a misplaced } in an #ifdev clause in element.hh. I will push a PR to fix it.

maxvanliefde:
Due to a misplaced }, compilation fails when FastClick is configured with statistics set to 2. 
This fixes #409.

### DIFF
--- a/include/click/element.hh
+++ b/include/click/element.hh
@@ -812,8 +812,8 @@ Element::Port::push(Packet* p) const
 # else
     _e->push(_port, p);
 # endif
-    }
 #endif
+    }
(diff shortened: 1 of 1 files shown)


tbarbette:
The main branch should be fixed now, can you rebase so the CI runs again on an healthy basis?

maxvanliefde:
Checks seem to pass, is it okay now?

PR #412, opened 2025-03-19
Refactor ControlSocket & HttpServer elements

Naya-01:
- Added 'MHD_OPTION_LISTENING_ADDRESS_REUSE' to allow the reuse of the listening address for the HTTPserver.
- replace 'SO_REUSEADD' with 'SO_REUSEPORT' in ControlSocket to allow reuse of the same port.


tbarbette:
I think you should find a way to send the elementmap.xml through a handler. Having to find it locally will be a highly limiting factor.

Naya-01:
I've just added a new handler for the elementmap.xml and it works quite well

PR #414, opened 2025-03-28
Commented out click_chatter("take!") in CheckIP6Header::simple_action

kkovacs:
Commented out click_chatter("take!") in CheckIP6Header::simple_action, because it's a too central place, has no huge value, and spams the log full. :)

Keep up the good work, Tom! Thank you for all your time you put into this project.


tbarbette:
I think that situation should be avoided though, that's why there is a message. But if you have a lot packets that arrive too big at the input, then there's not much choices indeed... The best way would be to print it 10 times and then stop, as done in some places.

kkovacs:
Hmm, you're right, limiting to 10 would be nice. And then maybe changing the message to a little bit more descriptive?

PR #415, opened 2025-04-08
Make elementmap.xml path resolution dynamic

Naya-01:
Make elementmap.xml path resolution dynamic by using CLICK_DATADIR variable

### DIFF
--- a/lib/router.cc
+++ b/lib/router.cc
@@ -35,6 +35,7 @@
 #include <click/notifier.hh>
 #include <click/nameinfo.hh>
 #include <click/bighashmap_arena.hh>
+#include <click/pathvars.h>
 #if HAVE_DPDK_PACKET_POOL || CLICK_PACKET_USE_DPDK
 #include <click/dpdkdevice.hh>
 #endif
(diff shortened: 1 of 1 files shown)


tbarbette:
Try multiple path in an if exists chain. If it's not in datadir, then /usr/local/... And also verify it works without doing make install and running from bin/click directly, because we always do that.

Naya-01:
how do you change the location of elementmap.xml without 'make install'?

tbarbette:
Can you rebase? The pipeline should be fixed now.

Naya-01:
A sync the repo all test passed but after the rebase, the test failed

tbarbette:
Can you rebase?

Naya-01:
> Can you rebase?

Done! Let me know if it's good like this or not.

tbarbette:
Maybe you forgot to fetch before the rebase? You're still 4 commits behind.

Naya-01:
it's normal because the “hash” of the commits has changed with the rebase but if you look at the history you can see that the commits are still there.

tbarbette:
You have modifications from me in your branch, which is not normal.
You have to:
```
git remote add tbarbette git@github.com:tbarbette/fastclick.git
git fetch --all
git rebase tbarbette/main
[skip commits you duplicated]
git push --force
```

tbarbette:
Thanks!

Issue #162, opened 2019-01-18
An Error in fastclick installation


signet68:
Hi all,

I have tried to install the fastclick on ubuntu 16.04 (and even on 14.04) according to the fastclick manual. 
I run the ./configure command in user mode:
./configure --enable-userlevel --disable-linuxmodule --enable-user-multithread
however, I have a problem with running make command
![screenshot from 2019-01-17 18-28-28](https://user-images.githubusercontent.com/46813150/51382057-03a1f480-1b2b-11e9-97a0-acbc3f62cec5.png)

Do you think that I should install an extra library?

tbarbette:
You are in a VM right? What is the architecture? Can you show the output of cat /proc/cpuinfo ?

signet68:
Thanks to your response.

No, I installed ubuntu 32bit directly on my laptop.
I attached a txt file as output of  cat /proc/cpuinfo.

[k.txt](https://github.com/tbarbette/fastclick/files/2772952/k.txt)

tbarbette:
What is your gcc/g++ version?

(PS : why a 32bit version? You have a 64bit processor, it will slow down the system for no reason?).

signet68:
![screenshot from 2019-01-18 17-29-19](https://user-images.githubusercontent.com/46813150/51391546-9c466d80-1b47-11e9-8e67-e59534c448dc.png)

I had installed the ubuntu18.04 64bit, but my system was really slow!

tbarbette:
The error is because of the 32bit system (thanks @aliireza ). I will try to fix it at the occasion. Your system being slow with 64bit is probably because of some other problem. Maybe try 16.04 64 bit?

signet68:
Thanks,
I could properly install the fastclick on ubuntu 16.04 64bit.
Already I installed Ubuntu 18.04, however, I had some problems with it.

tbarbette:
If someone has time to fix the issues on 32bit (simply some operations that are not available here and there) I would appreciate the help

dganeshk:
@tbarbette  able to compile it  on  18.04 32 bit ubuntu system. will push more PRs for other api as well in coming days. .

PR #423, opened 2025-09-02
[#162] Fix 64bit atomic api issues for 32bit system-part2

dganeshk:
*atomic apis are using only 64 bit assembly instructions([add,sub,..]q which are not supported in 32bit systems and compilation errors observed due to that. Fixing the assembly instruction for the api's add and subtract operator.


dganeshk:
@tbarbette  have pushed another PR with that all apis are completed,  All test cases in atomic.cc are passing with the changes. will push the atomic.cc test modification in a separate PR which is common for both 64 bit and 32 bit platforms.

dganeshk:
> Could you do a unique PR so I can try all at once?

@tbarbette  have raised PR with all changes in the below PR. 
https://github.com/tbarbette/fastclick/pull/431

PR #431, opened 2026-01-03
[#162] Fix 64bit atomic api issues for 32bit system

dganeshk:
*atomic apis are using only 64 bit assembly instructions([add,sub,..]q which are not supported in 32bit systems and compilation errors observed due to that. Fixing the assembly instruction for the apis.

*update test_64 to handle the 64bit test cases.


dganeshk:
attaching the test result of the ut in 32 bit system. 
image

